PR #11, opened 2017-04-05
Update the contributors command to "contributors.add"

CompuIves:
I noticed that `npm start "addContributor CompuIves"` didn't work, but `npm start "contributors.add"` does. This updates the documentation for it.


kentcdodds:
Ah! Thanks! Hard to keep things updated!

PR #22, opened 2017-04-06
WIP: feat(core): Add jsx style components to glamorous export

### LINKED ISSUE #20: Support jsxstyle components
https://github.com/smyte/jsxstyle experts several components. Glamorous exports element components that resemble the API. In addition to those element components, I'd like to also export the jsxstyle components because they have merit and make it more straightforward for people used to that API. We could learn from what glamor did to support this: https://github.com/threepointone/glamor/blob/master/src/jsxstyle.js

I expect this would just add a handful of lines of code...

paulmolluzzo:
This uses the components exported from `glamor/jsxstyle` and assigns them to the `glamorous` object
already exported as additional components.

**What**:
This is an attempt to fix #20 by including the jsxstyle components that `glamor` exports.

**Why**:
Suggested by @kentcdodds to ease developers into using the `glamorous` API.

**How**:
I'm importing the `glamor/jsxstyle` and using an array of their component names to extend  `glamorous` with these components, nearly identically to the way `glamorous` adds support for components like `glamorous.Div`

Currently this is not really working for me, and I'm not sure if I'm approaching this entirely incorrectly or what. If I "manually" modify `glamorous` in a project to do exactly the same things that I'm doing in this commit it works, but it seems to fail. I might be overlooking something, so if @kentcdodds or anyone has advice please feel free to comment.

I'm testing by importing this updated build into the [`with-glamorous` example for Next](https://github.com/zeit/next.js/tree/master/examples/with-glamorous).

The error I get now is:

```
Cannot read property 'comp' of undefined
TypeError: Cannot read property 'comp' of undefined
    at getGlamorousComponentMetadata (/Users/paul/open-source/glamorous/dist/glamorous.cjs.js:6233:28)
    at Function.glamorousComponentFactory [as view] (/Users/paul/open-source/glamorous/dist/glamorous.cjs.js:6222:39)
    at /Users/paul/open-source/glamorous/dist/glamorous.cjs.js:6310:37
    at Array.reduce (native)
    at Object.<anonymous> (/Users/paul/open-source/glamorous/dist/glamorous.cjs.js:6308:48)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```

If there's any feedback or direction to help get this done I'd be super grateful! 🙏

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,10 +4,9 @@
  */
 import React from 'react'
 import {css, styleSheet} from 'glamor'
-import * as jsxStyleComponents from 'glamor/jsxstyle'
 import shouldForwardProperty from './should-forward-property'
 import domElements from './dom-elements'
-import jsxStyleComponentTags from './jsxstyle-components'
+import jsxStyleComponents from './jsxstyle-components'
 
 const {PropTypes} = React
 
@@ -130,23 +129,18 @@ Object.assign(
 /*
  * This creates a glamorousComponentFactory for every jsxstyle element so you
  * can simply do:
- * const GreenBlock = glamorous.block({
+ * const GreenBlock = glamorous.display.block({
  *   backgroundColor: 'green',
  *   padding: 20,
  * })
  * <GreenBlock>Click Me!</GreenBlock>
  */
-Object.assign(
-  glamorous,
-  jsxStyleComponentTags.reduce(
-    (getters, tag) => {
-      const capitalTag = capitalize(tag)
-      const jsxStyleComponent = jsxStyleComponents[capitalTag]
-      getters[tag] = glamorous(jsxStyleComponent)
-      return getters
-    },
-    {},
-  ),
+glamorous.display = jsxStyleComponents.reduce(
+  (getters, {name, tag, styles}) => {
+    getters[name] = glamorous[tag](styles)
+    return getters
+  },
+  {},
 )
 
 /*
@@ -176,18 +170,18 @@ Object.assign(
 /*
  * This creates glamorous components matching components produced by jsxstyle
  * simply do:
- * <glamorous.Block>
+ * <glamorous.display.Block>
  *   I'm a block!
- * </glamorous.Block>
+ * </glamorous.display.Block>
  */
 Object.assign(
-  glamorous,
-  jsxStyleComponentTags.reduce(
-    (comps, tag) => {
-      const capitalTag = capitalize(tag)
-      comps[capitalTag] = glamorous[tag]()
-      comps[capitalTag].displayName = `glamorous.${capitalTag}`
-      comps[capitalTag].propsAreCssOverrides = true
+  glamorous.display,
+  jsxStyleComponents.reduce(
+    (comps, {name, tag, styles}) => {
+      const capitalName = capitalize(name)
+      comps[capitalName] = glamorous[tag](styles)
+      comps[capitalName].displayName = `glamorous.${capitalName}`
+      comps[capitalName].propsAreCssOverrides = true
       return comps
     },
     {},
--- a/src/index.js
+++ b/src/index.js
@@ -4,8 +4,10 @@
  */
 import React from 'react'
 import {css, styleSheet} from 'glamor'
+import * as jsxStyleComponents from 'glamor/jsxstyle'
 import shouldForwardProperty from './should-forward-property'
 import domElements from './dom-elements'
+import jsxStyleComponentTags from './jsxstyle-components'
 
 const {PropTypes} = React
 
@@ -126,6 +128,28 @@ Object.assign(
 )
 
 /*
+ * This creates a glamorousComponentFactory for every jsxstyle element so you
+ * can simply do:
+ * const GreenBlock = glamorous.block({
+ *   backgroundColor: 'green',
+ *   padding: 20,
+ * })
+ * <GreenBlock>Click Me!</GreenBlock>
+ */
+Object.assign(
+  glamorous,
+  jsxStyleComponentTags.reduce(
+    (getters, tag) => {
+      const capitalTag = capitalize(tag)
+      const jsxStyleComponent = jsxStyleComponents[capitalTag]
+      getters[tag] = glamorous(jsxStyleComponent)
+      return getters
+    },
+    {},
+  ),
+)
+
+/*
  * This creates a glamorous component for each DOM element so you can
  * simply do:
  * <glamorous.Div
@@ -138,6 +162,27 @@ Object.assign(
 Object.assign(
   glamorous,
   domElements.reduce(
+    (comps, tag) => {
+      const capitalTag = capitalize(tag)
+      comps[capitalTag] = glamorous[tag]()
+      comps[capitalTag].displayName = `glamorous.${capitalTag}`
+      comps[capitalTag].propsAreCssOverrides = true
+      return comps
+    },
+    {},
+  ),
+)
+
+/*
+ * This creates glamorous components matching components produced by jsxstyle
+ * simply do:
+ * <glamorous.Block>
+ *   I'm a block!
+ * </glamorous.Block>
+ */
+Object.assign(
+  glamorous,
+  jsxStyleComponentTags.reduce(
     (comps, tag) => {
       const capitalTag = capitalize(tag)
       comps[capitalTag] = glamorous[tag]()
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -4,8 +4,10 @@
  */
 import React from 'react'
 import {css, styleSheet} from 'glamor'
+import * as jsxStyleComponents from 'glamor/jsxstyle'
```
kentcdodds: I'd rather not import this from `glamor/jsxstyle` especially considering that glamor is dropping this in favor of `glamorous` in 3.0 ;-) We should construct the components ourselves. And actually it should be fairly simple and straightforward to do that.

src/jsxstyle-components.js
```diff
@@ -0,0 +1,9 @@
+// Mostly copied from:
+// https://github.com/threepointone/glamor/blob/
+// 5e7d988211330b8e2fca5bb8da78e35051444efd/src/jsxstyle.js
+/*
+ * This is used to generate the helper functions
+ * like: glamorous.block and the <glamorous.Block /> elements
+ */
+
+export default ['view', 'block', 'inlineBlock', 'flex', 'row', 'column']
```
kentcdodds: if we weren't using rollup, I'd prefer to have this in the same file to save on module closures, but thanks to rollup's module hoisting we should be good here! 🎉

src/index.js
```diff
@@ -126,6 +128,28 @@ Object.assign(
...
+ *   backgroundColor: 'green',
+ *   padding: 20,
+ * })
+ * <GreenBlock>Click Me!</GreenBlock>
+ */
+Object.assign(
+  glamorous,
+  jsxStyleComponentTags.reduce(
+    (getters, tag) => {
+      const capitalTag = capitalize(tag)
```
kentcdodds: I think this is actually capitalizing the tag which we wouldn't want to do in this case. However, I'm unsure I want to support `glamorous.block` and `glamorous.flex` etc... Hmmm... What do you think?

src/jsxstyle-components.js
kentcdodds: Actually, because we're going to be constructing things ourselves (rather than using glamor's implementation) perhaps this should be more of a config:

```js
export default [
  {name: 'view', tag: 'div'},
  {name: 'inlineBlock', tag: 'div', styles: {display: 'inline-block'}},
  // etc...
]
```

Then in `index.js` we can reduce these to glamorousFactories with the styles pre-applied:

```
(comps, {name, tag, styles}) => {
  const capitalName = capitalize(name)
  comps[capitalName] = glamorous[tag](styles)
  // ..etc
}
```

I'm also thinking that it may be better to not put them directly on `glamorous` but instead scope them to a property on `glamorous`. Maybe `glamorous.display`? Because they're all display-specific? So you'd do:

```javascript
const {View, InlineBlock} = glamorous.display

// use them here
```

Thoughts?

src/index.js
paulmolluzzo: Fair point.

src/index.js
paulmolluzzo: The other components are capitalized, so I followed suit. For that reason I say capitalize, but I don't have a strong opinion either way.

kentcdodds: What I mean is, in the code block example above it shows: `glamorous.block`, but here it's capitalizing the tag, so what you'll actually have to do is: `glamorous.Block`, which will actually get overwritten in the next `Object.assign` anyway.

In any case, I'm unsure I want to expose this. And if we do I think I'd want to expose it on `glamorous.display` as discussed in the other comment.

src/jsxstyle-components.js
paulmolluzzo: > Actually, because we're going to be constructing things ourselves (rather than using glamor's implementation) perhaps this should be more of a config:

I had thought about a solution like this at one point, and I like this approach a little better even.


> I'm also thinking that it may be better to not put them directly on glamorous but instead scope them to a property on glamorous. Maybe glamorous.display? Because they're all display-specific? So you'd do:

The other components like `a` and `div` are right on `glamorous` though. It's a little weird that some components would be scoped lower than others. You'd have to find where we shoved all these elements. 😄 

If we're going to put these on a `glamorous.display` perhaps we should put them all there?

kentcdodds: I think the assumption is that DOM elements are accessible directly on `glamorous`. If we start shoving other things on `glamorous` would that get confusing? Maybe not I guess.

paulmolluzzo: > I think the assumption is that DOM elements are accessible directly on `glamorous`. If we start shoving other things on `glamorous` would that get confusing? Maybe not I guess.

It'd get a little cluttered. I like your idea of `.display`, I just wonder if all the other elements might be better off moved there as well.

`import { A, Flex, Row, Column } from glamorous.display`

kentcdodds: Ah, that gives me an idea, what if we did:

```javascript
import { Block, Flex, Row, Column } from 'glamorous/display'
```

Might be a little bit of a challenge for the UMD build though 🤔  We'd either have to have an entirely different file you'd have to load, or still tack it onto `glamorous.display`...

Hmmm... I think that I just like this the best:

```javascript
const {A, Img, Div} = glamorous
const {Block, Flex, Row, Column} = glamorous.display
const MyDiv = glamorous.div({/* styles */})
```

It suffers from 0 UMD issues, and it makes a clear separation between the original use-case of elements and the "display" use case of display elements.

paulmolluzzo: Update coming in a moment that does what you like best. 😉

paulmolluzzo:
I pushed an update that addresses all the notes so far, but I'm welcome to hear more thoughts and make corrections as needed.

You can test locally with something like:

```js

const { Flex } = glamorous.display

export default () => <Flex>Test View</Flex>
```

src/index.js
```diff
@@ -126,6 +127,23 @@ Object.assign(
...
 /*
+ * This creates a glamorousComponentFactory for every jsxstyle element so you
+ * can simply do:
+ * const GreenBlock = glamorous.display.block({
+ *   backgroundColor: 'green',
+ *   padding: 20,
+ * })
+ * <GreenBlock>Click Me!</GreenBlock>
+ */
+glamorous.display = jsxStyleComponents.reduce(
```
paulmolluzzo: @kentcdodds Is there a better way or place to do this?

kentcdodds:
I've actually been thinking about this a little further and I'm concerned that by adding this is adding more complexity without providing as much benefit. Or maybe that it's out of scope for the purpose of this package.

With that in mind, maybe it would be better to make another package (`glamorous-jsxstyle`?) that exposes this. So you could do:

```javascript
import {Flow, Row, InlineBlock} from 'glamorous-jsxstyle'
```

I feel much better about that I think... What are your thoughts?

paulmolluzzo:
Yeah, I see your point, and adding these additional display components starts confusing the purpose of this package. I don't view `glamorous` as a component library, even though it provides DOM elements as components.

The pattern you suggested makes sense. Isn't this all it would be?

```js
import glamorous from 'glamorous';

export default () => {
  View: glamorous.div()
  Block: glamorous.div({display: 'block'})
  InlineBlock: glamorous.div({display: 'inline-block'})
  Flex: glamorous.div({display: 'flex'})
  Row: glamorous.div({display: 'flex'; flexDirection: 'row'})
  Column: glamorous.div({display: 'flex'; flexDirection: 'column'})
}
```

kentcdodds:
Yep! That's all it'd be! Do you wanna make it?! We could then link to it in the README of this project and add an example maybe :smile:

kentcdodds:
I'll go ahead and close this as it appears we're headed in a better direction. Thanks!

kentcdodds:
Actually, I just updated your comment to be more correct :+1:

paulmolluzzo:
I can make it, sure! Might not get to it for a few hours though.

kentcdodds:
Take your time. The world can wait :)

paulmolluzzo:
@kentcdodds https://www.npmjs.com/package/glamorous-jsxstyle

kentcdodds:
Dude, that's perfect! Way to go!

paulmolluzzo:
Thanks for the help!

PR #24, opened 2017-04-06
feat(lib): Theming support

vesparny:
**What**:
Theming support 

**Why**:
It is part of the project roadmap

**How**:
A new component has been added:
`ThemeProvider`
It takes a `theme` props that will be propagated down the tree via context.

Context updates will also be propagated to children that return false from `shouldComponentUpdate`.
This has been achieved using an event emitter implemented with [mitt](https://github.com/developit/mitt) and inspired by https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076

ThemeProviders can also be nested.
Inner components will merge outer theme with own.

I had to change `GlamorousComponent` to a class in order to access lifecycle methods.

**TODO**:
* Adding unit tests (I have no experience with jest, but I'd love to follow your guidance)
* There is one test that currently fails (see previous point)
* Documentation
* Anything else ?

An example: 
```jsx
import ReactDOM from 'react-dom'
import React, {Component} from 'react'
import glamorous, {ThemeProvider} from 'glamorous'

const Text = glamorous.div({
  color: 'blue',
},
(props, theme) => {
  return {
    fontSize: theme.fontSize * 100
  }
})

class Root extends Component {
  constructor () {
    super()
    this.state = {
      fontSize: 1
    }
    this.tick = this.tick.bind(this)
  }

  componentDidMount () {
    this.tick()
  }
  tick () {
    setInterval(() => this.setState({fontSize: Math.random()}), 2000)
  }


  render () {
    return (
      <ThemeProvider theme={{fontSize: this.state.fontSize}}>
          <Text>
            <ThemeProvider theme={{anotherThemeProperty: 100}}>
              <div>
                <Text>My font size will change</Text>
                <Pure>
                  <Text>My font size will still change, even if I'm inside a pure component</Text>
                </Pure>
              </div>
            </ThemeProvider>
          </Text>
      </ThemeProvider>
    )
  }
}

class Pure extends Component {
  shouldComponentUpdate() {
    return false
  }

  render() {
    return <Text>{this.props.children}</Text>
  }
}

ReactDOM.render(
  <Root />,
  document.getElementById('root')
)
```


src/index.js
```diff
@@ -40,40 +41,72 @@ function glamorous(comp, {rootEl, displayName} = {}) {
...
-      return React.createElement(GlamorousComponent.comp, {
-        className: fullClassName,
-        ...toForward,
-      })
-
-      function mapToPropsCall(glamorRules) {
-        if (typeof glamorRules === 'function') {
-          return glamorRules(props)
+    class GlamorousComponent extends Component {
+      constructor(props, context) {
```
kentcdodds: If we use public class fields, then we wont even need this constructor. Could just be:

```javascript
class GlamorousComponent extends Component {
  state = {theme: null}
  setTheme = () => this.setState({theme})
  // etc.
}
```

src/index.js
```diff
@@ -40,40 +41,72 @@ function glamorous(comp, {rootEl, displayName} = {}) {
...
+        this.setTheme = this.setTheme.bind(this)
+      }
+
+      setTheme(theme) {
+        this.setState({
+          theme,
+        })
+      }
+
+      componentWillMount() {
```
kentcdodds: I prefer using `componentDidMount` for stuff like this. Is there a reason you used `componentWillMount`. From [the react docs](https://facebook.github.io/react/docs/react-component.html#componentwillmount):

> `componentWillMount()` is invoked immediately before mounting occurs. It is called before render(), therefore setting state in this method will not trigger a re-rendering. Avoid introducing any side-effects or subscriptions in this method.

src/index.js
```diff
@@ -40,40 +41,72 @@ function glamorous(comp, {rootEl, displayName} = {}) {
...
+          className: fullClassName,
+          ...toForward,
+        })
       }
     }
+
     GlamorousComponent.propTypes = {
       className: PropTypes.string,
       cssOverrides: PropTypes.object,
+      theme: PropTypes.object,
```
kentcdodds: So if I provide a `theme` prop to a `GlamorousComponent`, will that override the context for that component? Is that the purpose of exposing this prop? Sorry if I'm not following that bit.

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+        subscribe: this.subscribe,
+      },
+    }
+  }
+
+  setOuterTheme(theme) {
+    this.outerTheme = theme
+  }
+
+  componentWillMount() {
```
kentcdodds: Same thing here, I'd prefer to use `componentDidMount`

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+    // merge this theme with the outer theme if present
```
kentcdodds: This is pretty cool. Neat feature we should definitely highlight in the docs :+1:

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
+import React, {PropTypes, Component} from 'react'
```
kentcdodds: I prefer to keep my files lower case and snake-cased (so this would be `theme-provider.js`. I should probably have a lint rule for that. But the reason is git has a hard time renaming files when it's purly case-changes and some filesystems are not case sensitive so it can lead to confusion.

kentcdodds:
Also, based on the travis build, it looks like we're going to need to add `babel-plugin-external-helpers` to our `devDependencies` and add it to the [`rollup.config.js`](https://github.com/paypal/glamorous/blob/f75fd4556edf826c927e6ffba20a7f950229b133/rollup.config.js#L28-L32).

package.json
```diff
@@ -24,10 +24,12 @@
...
   "license": "MIT",
-  "dependencies": {},
+  "dependencies": {
+    "mitt": "^1.1.0"
+  },
   "peerDependencies": {
-    "react": ">=14",
-    "glamor": ">=2"
+    "glamor": ">=2",
+    "react": ">=14"
```
vesparny: yup

src/ThemeProvider.js
developit: Kent we are the only two people on Earth who do this. Go team lowercase-with-dashes!

src/ThemeProvider.js
developit: Yup, otherwise you'll have dangling handlers during SSR

src/ThemeProvider.js
developit: A+ on this decision.

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+
+  // create them, by merging with outer theme, if present
+  getTheme(passedTheme) {
+    const theme = passedTheme || this.props.theme
+    return {...this.outerTheme, ...theme}
+  }
+
+  getChildContext() {
+    return {
+      ...this.context,
```
developit: No need to mix the existing context in here, that's actually already done for you.

kentcdodds: That's a great piece of insight I never would have known. Thanks!

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+      this.publish(this.getTheme(nextProps.theme))
+    }
+  }
+
+  componentWillUnmount() {
+    this.unsubscribe && this.unsubscribe()
+  }
+
+  render() {
+    return React.Children.only(this.props.children)
```
kentcdodds: Oh cool! Thanks for that :)

Yeah, the `ThemeProvider` should never be used like: `<ThemeProvider>hello world</ThemeProvider>`, but it _could_ be used like:

```javascript
<ThemeProvider>
  {shouldRender ? <Stuff /> : null}
</ThemeProvider>
```

Especially with the nesting capabilities this PR enables (can't think of a use case for the root-level provider). Am I off base here or is that something that would blow up with this code?

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+    emitter.emit(CHANNEL, this.getTheme())
+
+    return function unsubscribe() {
+      emitter.off(CHANNEL, subscription)
+    }
+  }
+
+  // create them, by merging with outer theme, if present
+  getTheme(passedTheme) {
+    const theme = passedTheme || this.props.theme
```
kwelch: Only difference would be if you wanted to allow null or now. Defaulting will not catch null. That is preference based. Not sure how spreading works on null.

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
+import React, {PropTypes, Component} from 'react'
+import mitt from 'mitt'
+
+export const CHANNEL = '__glamorous__'
```
kwelch: Would you want this to be a symbol to ensure uniqueness?

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * This is a component which will provide a theme to the entire tree
+ * via context and event listener (because pure components block conext updates)
+ * @param {Object} props the props for the component.
+ * @param {Object} context the context for the component.
+ * @return {ReactElement} React.createElement
+ */
+class ThemeProvider extends Component {
+  constructor(props, context) {
+    super(props, context)
```
aweary: You don't have to pass `props` and `context` to `super` if you're not accessing them on the instance (`this.props`) in the `constructor`

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+  getChildContext() {
+    return {
+      ...this.context,
+      [CHANNEL]: {
+        subscribe: this.subscribe,
+      },
+    }
+  }
+
+  setOuterTheme(theme) {
```
kwelch: Should you emit on set? CompA sets theme, you should be able to pick it up in CompB.

src/ThemeProvider.js
kentcdodds: Yeah, I'd considered that as well. `{...null}` probably doesn't work out too well. Maybe we should leave this as-is? Is there a situation where the theme could be `null`?

kentcdodds: Who am I kidding. This is JavaScript OF COURSE. Maybe we should do:

```javascript
const theme = passedTheme || this.props.theme || {}
```

src/index.js
aweary: The advantage of using `componentDidMount` here is that the `setState` call in `setTheme` will not cause a re-render after the initial render.

src/ThemeProvider.js
aweary: Correct, `null` is not a valid React element so it would throw a (misleading) error. [Here's a JSFiddle](https://jsfiddle.net/0fLnnyt2/) showing that behavior. You can probably just add a null check before `React.Children.only`, or do your own validation

developit: FWIW I'd stick with `Children.only()` versus a direct return here, but I'm biased - in preact `children` is always an Array, so having third party libs use `Children.only()` provides a nice way to normalize that behavior.

src/index.js
developit: Indeed I'd do the state-setting check in `componentWillMount()` and the subscription in `componentDidMount()`.  That gives you a good initial synchronous render and avoids listening when doing SSR.

src/ThemeProvider.js
browniefed: WOW. I already spread context in. I assume this won't change in future context changes that haven't been decided on by React core?

developit: Would break every React app, so I'm thinking it's safe :P

src/ThemeProvider.js
kentcdodds: I'd considered that. Maybe we could do:

```javascript
export const CHANNEL = typeof Symbol !== 'undefined' ? Symbol('__glamorous__') : '__glamorous__'
```

kentcdodds:
@kwelch 

> a HoC that wraps the call to context

Could you give a more clear example of what you mean? I don't do a whole lot with Providers, so this is a little foreign to me.

src/ThemeProvider.js
aweary: Well `Children.only` doesn't do anything except validation, it will always return the same value you pass in. It's a very confusing API and doesn't really do what it says it does 😄

But I agree, it's worth keeping for validation (we can fix the error message so it's less confusing) and just do an additional `null` check beforehand.

browniefed: Should this be removed when Fiber lands since multiple returns will be supported?

kentcdodds: I'm good with keeping it as-is :) 🙇

src/index.js
vesparny: @developit The problem is that if we do the subscription in `componentDidMount`, the theme won't be available in `componentWillMount`.
Theme is emitted by mitt for the first time on after invoking `subscribe`.

I see your point, just wondering how to manage it

src/ThemeProvider.js
aweary: The `CHANNEL` key is used in `contextTypes`, I'm not positive that React currently supports using Symbols for keys in context. It would be worth verifying before making this change.

src/ThemeProvider.js
aweary: @browniefed The `Children.only` API should probably be adjusted instead of requiring consumers to remove it. Something that still needs to be figured out I think 😃

src/ThemeProvider.js
kentcdodds: Ah, good point. Unit tests will help verify this :+1:

kwelch:
This PR exposes functions through context so to use them the implementor will have to import your `CHANNEL` and request access to context from react. 

If you make an HoC you can abstract the need to use context from the user. 

React-broadcast by Michael Jackson does this will the component subscriber​. The provider in that library is broadcaster. https://github.com/mjackson/react-broadcast

This is a common pattern to help isolate changes if/when context goes away.

kentcdodds:
Thanks for the explination @kwelch. I'm trying to think of a situation where someone would want the theme outside of a glamorous component though... I suppose it's possible. But I don't think I mind making it a little harder for them to get it because it _should_ only be used for `glamorous` components (which are created via a semi-HoC anyway).

src/theme-provider.js
```diff
@@ -0,0 +1,85 @@
+import React, {PropTypes, Component} from 'react'
+import mitt from 'mitt'
+
+export const CHANNEL = '__glamorous__'
```
aweary: @kentcdodds @kwelch (continuing from the now collapsed conversation)

I checked and [using Symbols in `context`](https://jsfiddle.net/v2fb93bm/) does not work, so this should remain a string.

kwelch:
Totally makes sense. I think that part of I was missing from bring on a phone. Makes sense. 👍

vesparny:
@kentcdodds 
I pushed few changes addressing your suggestions.

@developit 
I think I understood what you meant by 
> indeed I'd do the state-setting check in componentWillMount() and the subscription in componentDidMount(). That gives you a good initial synchronous render and avoids listening when doing SSR.

I've implemented a subscriber that has been basically copy/pasted from react-broadcast.
Thanks to this we are now able to setup the subscription in `componentDidMount` and settting the state in `componentWillMount`

src/ThemeProvider.js
developit: I'd actually recommend against the Symbol approach, having just removed that exact setup from Preact.  Most people will end up getting a Symbol polyfill in their bundle (which includes polyfillying `typeof`, which is slow).  Just use an underscore key like you have there.

FWIW symbols are supported in context, just unnecessary. Also fun caveat: with the string you can have 2 copies of this lib and they will talk properly. With the Symbol, if you bundle twice and get 2 polyfills, they may not have the same generated key.

developit:
Regarding mitt - generally the way you'd do that is by storing your state in context (on the EventEmitter) as a simple property, rather than firing an event to obtain it.  This allows you to access the state synchronously during componentWillMount.

vesparny:
@developit 
I guess you mean something like
```js
import mitt from 'mitt'

export default function createBroadcast(initialState, CHANNEL) {
  const emitter = mitt()
  let currentState = initialState

  function getState() {
    return currentState
  }

  function setState(state) {
    currentState = state
    emitter.emit(CHANNEL, currentState)
  }

  function subscribe(listener) {
    emitter.on(CHANNEL, listener)

    return function unsubscribe() {
      emitter.off(CHANNEL, listener)
    }
  }

  return {getState, setState, subscribe}
}

```

I'd go with this solution, this way we do not need to maintain the pub/sub code. Mitt is so tiny that it is like it's not even there :)

@kentcdodds what do you think?

src/index.js
vesparny: This is exactly what I had in mind.
In case a prop `theme` is provided, it takes precedence over context.
I would not even merge with context in this case.

src/ThemeProvider.js
javivelasco: You are not the only two people. Check react-toolbox 😂

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+ * @return {ReactElement} React.createElement
+ */
+class ThemeProvider extends Component {
+  constructor(props, context) {
+    super(props, context)
+    this.getTheme = this.getTheme.bind(this)
+    this.setOuterTheme = this.setOuterTheme.bind(this)
+    this.emitter = mitt()
+    this.publish = this.emitter.emit.bind(this, CHANNEL)
+    this.subscribe = this.subscribe.bind(this)
```
javivelasco: I was about to suggest the same thing

src/ThemeProvider.js
```diff
@@ -0,0 +1,89 @@
...
+  // create them, by merging with outer theme, if present
+  getTheme(passedTheme) {
+    const theme = passedTheme || this.props.theme
+    return {...this.outerTheme, ...theme}
+  }
+
+  getChildContext() {
+    return {
+      ...this.context,
+      [CHANNEL]: {
```
javivelasco: What about allowing the user to define the `CHANNEL`? This way one could write two completely separated providers feeding two different themes. Using the same channel would do the merging thing

developit:
@vesparny yup, that looks good.

vesparny:
@developit awesome.
I've extracted it into its own package as I often use that pattern.

https://github.com/vesparny/brcast

I think we can include it in this PR.

src/theme-provider.js
```diff
@@ -0,0 +1,74 @@
...
+  componentWillMount() {
+    // set broadcast state by mergin outer theme with own
+    if (this.context[CHANNEL]) {
+      this.broadcast.setState(this.context[CHANNEL].getState())
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.theme !== nextProps.theme) {
+      this.broadcast.setState(this.getTheme(this.getTheme(nextProps.theme)))
```
kwelch: Was the double call to `getTheme` intentional?

src/ThemeProvider.js
vesparny: I do not know. Personally I would not give this freedom to users.
To me a theme is something global and I've never felt the need of having more than one in same app. (but maybe I've never ran into that scenario 😱)

Do you think it would be a nice addition considering we already provide merging capabilities? 
Should not be complex to implement.

src/theme-provider.js
vesparny: @kwelch nope it's a silly mistake 🙄
Thanks for catching it.

developit:
@vesparny one suggestion - make the channel an argument to the constructor function so glamorous can use `__glamorous__`

rollup.config.js
```diff
@@ -14,7 +14,7 @@ export default {
         {dest: 'dist/glamorous.es.js', format: 'es'},
         {dest: 'dist/glamorous.cjs.js', format: 'cjs'},
   ],
-  exports: 'default',
+  exports: 'named',
```
kentcdodds: I don't know rollup very well, but we'll wanna make sure that all the `dist/` files still work the same after these changes. Thanks!

src/index.js
```diff
@@ -35,43 +36,77 @@ function glamorous(comp, {rootEl, displayName} = {}) {
...
-          return glamorRules(props)
+    class GlamorousComponent extends Component {
+      state = {theme: null}
+      setTheme = theme => this.setState({theme})
+
+      componentWillMount() {
+        const {theme} = this.props
+        if (this.context[CHANNEL]) {
+          // if a theme is provided via props, it takes precedence over context
+          this.setTheme(theme ? theme : this.context[CHANNEL].getState())
```
kentcdodds: If we want to support setting the `theme` to `null` (which we sorta do below), then this should probably be:

```javascript
this.setTheme(theme || this.context[CHANNEL].getState())
```

src/index.js
```diff
@@ -35,43 +36,77 @@ function glamorous(comp, {rootEl, displayName} = {}) {
...
-      })
-
-      function mapToPropsCall(glamorRules) {
-        if (typeof glamorRules === 'function') {
-          return glamorRules(props)
+    class GlamorousComponent extends Component {
+      state = {theme: null}
+      setTheme = theme => this.setState({theme})
+
+      componentWillMount() {
```
kentcdodds: Just to be clear, it's safe for us to use `componentWillMount` here because we can get the state immediately via the `.getState()` call (we don't have to wait for the emitter to emit the first theme state. Correct?

src/index.js
kentcdodds: That sounds like a good decision :+1: Thanks!

rollup.config.js
vesparny: It's the first time I use it as well.
As far as in understood, 'named' is necessary when you export default plus named

src/index.js
vesparny: exactly

rollup.config.js
developit: I believe this means your commonjs exports object will be frozen. You might want to use `default` or [prune via memory plugin](https://github.com/developit/preact-compat/blob/master/rollup.config.js#L27-L30).

vesparny:
At the moment one test is failing but I could not understand why.

I will look into jest next week because I've never used it before. Won't probably make it during the weekend.

Some help would be very much appreciated.
@kentcdoods do you know if anyone is willing to help.
If so, how it works in those cases? Should I give access to my fork to other contributors?

kentcdodds:
Feel free to take your time. I believe I can make updates to your PR as well so if I find the time I may be able to help out a little bit too. Thanks again for all your help (and thanks for the PR reviewers as well).

kwelch:
If someone wants to help they should be able in theory to fork and PR into your branch(vesparny:patch-1) right?

rollup.config.js
kentcdodds: I think that I want the UMD global to be the `glamorous` function with `glamorous.ThemeProvider`. I don't want `glamorous.default`. Is this possible? Same thing with the CJS export.

vesparny: I saw somewhere in the rollup docs that `named` has to be used when the lib exports something like we do.

```js
export default glamorous
export {ThemeProvider}
```
Isn't it correct?

vesparny:
First time I use jest, it's been fun 👍

rollup.config.js
vesparny: @kentcdodds I think we can do 
```js
Object.assign(
  glamorous,
  ThemeProvider,
)
```

import {Div, ThemeProvider} from 'glamorous'

kentcdodds:
I'll give this a solid look tomorrow probably. Just thought I'd let you know that I think your branch is out of date with master. Could you update that? Thanks so much for this work. It's fantastic!

vesparny:
rebased. not it is up to date with master.

rollup.config.js
vesparny: Using `named` in rollup config we can have something like:

```js
import glamorous, {ThemeProvider} from 'glamorous'
// like react itself
import React, {Component} from 'react'

```

if you want to have a single default export we will have something like

```js

// in the lib 
Object.assign(
  glamorous,
  {ThemeProvider},
)

// consuming the lib
const {ThemeProvider} = glamorous
```

In the end it does not change that much.
@kentcdodds your choice :)

README.md
```diff
@@ -282,12 +282,63 @@ and `A` components which make use of `src` and `href` props).
...
 One other tip... This totally works:
 
-```javascript
+```jsx
 <glamorous.Div color="blue">
   JSX is pretty wild!
 </glamorous.Div>
 ```
 
+### Theming
```
kentcdodds: Great docs! Once this is merged, Could you add a link to a CodeSandbox so people can play around with it? Could be based on [the help one](http://kcd.im/glamorous-help)

other/ROADMAP.md
```diff
@@ -6,8 +6,6 @@ We haven't filled this out yet though. Care to help? See `CONTRIBUTING.md`
 
 ## Want to do
 
-- Support theming via some sort of `ThemeProvider`
```
kentcdodds: :tada:

rollup.config.js
kentcdodds: Ok, cool, so this is what I want to be able to do:

```javascript
// when using ESM
import glamorous, {ThemeProvider} from 'glamorous'

// when using CJS
const glamorous = require('glamorous')
const {ThemeProvider} = glamorous

// when using globals
const glamorous = window.glamorous
const {ThemeProvider} = glamorous
```

I'm pretty sure this can be accomplished if we do:

```javascript
Object.assign(glamorous, {ThemeProvider})

export default glamorous
export {ThemeProvider}
```

It's kinda monkey-patching, but hey, this is JavaScript :smile:

src/__tests__/__snapshots__/index.js.snap
```diff
@@ -35,6 +35,46 @@ exports[`can use pre-glamorous components with css attributes 1`] = `
...
 `;
 
+exports[`enzyme.mount 1`] = `
+.css-1fyph4i,
+[data-css-1fyph4i] {
+  color: red;
+  padding: 10px;
+}
+
+<glamorous(div)
```
kentcdodds: Looking forward to #29 so this is even better :smile:

src/__tests__/index.js
```diff
@@ -1,15 +1,25 @@
 /* eslint func-style:0 */
 import React from 'react'
 import * as glamor from 'glamor'
-import {render} from 'enzyme'
+import {render, mount} from 'enzyme'
 import * as jestGlamorReact from 'jest-glamor-react'
 import {oneLine} from 'common-tags'
+import toJson from 'enzyme-to-json'
```
kentcdodds: Because we use [the serializer](https://github.com/paypal/glamorous/blob/14970c1bb59d1408d3066d209dafa356b5ce3173/package.json#L97-L99) this shouldn't be necessary :tada:

src/__tests__/index.js
```diff
@@ -192,3 +204,46 @@ test('forwards props when the GlamorousComponent.rootEl is known', () => {
...
+  const Comp = glamorous.div(
+    {
+      color: 'red',
+    },
+    (props, theme) => ({padding: theme.padding}),
+  )
+  const wrapper = mount(<Comp theme={{padding: '10px'}} />)
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
+  wrapper.setProps({theme: {padding: '20px'}})
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
```
kentcdodds: As noted above, because we use [the serializer](https://github.com/paypal/glamorous/blob/14970c1bb59d1408d3066d209dafa356b5ce3173/package.json#L97-L99) this could simply be:

```javascript
expect(wrapper).toMatchSnapshotWithGlamor('enzyme.mount')
```

src/__tests__/index.js
```diff
@@ -146,7 +156,9 @@ test('forwards props when the GlamorousComponent.rootEl is known', () => {
   })()
   // no need to pass anything. This will just create be a no-op class,
   // no problem
-  const MyWrappedVersionMock = jest.fn(MyWrappedVersion)
```
kentcdodds: Interesting, is there a reason the original wasn't working?

src/__tests__/index.js
```diff
@@ -192,3 +204,46 @@ test('forwards props when the GlamorousComponent.rootEl is known', () => {
...
+
+test('passes an updated theme when theme prop changes', () => {
+  const Comp = glamorous.div(
+    {
+      color: 'red',
+    },
+    (props, theme) => ({padding: theme.padding}),
+  )
+  const wrapper = mount(<Comp theme={{padding: '10px'}} />)
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
```
kentcdodds: Could we give both of these a more informative name rather than simply `enzyme.mount`. Now that we have several tests in this file it may be useful to have it be: `'with padding 10px theme'` and then `'with theme prop of padding 20px'` or something. That way reviewing the snapshots is easier.

src/__tests__/index.js
```diff
@@ -192,3 +204,46 @@ test('forwards props when the GlamorousComponent.rootEl is known', () => {
...
+  wrapper.setProps({theme: {padding: '20px'}})
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
+})
+
+test('cleans up theme subscription when unmounts', () => {
```
kentcdodds: Awesome test. Thank you :+1:

src/__tests__/index.js
```diff
@@ -192,3 +204,46 @@ test('forwards props when the GlamorousComponent.rootEl is known', () => {
...
+  const unsubscribe = jest.fn()
+  const context = getMockedContext(unsubscribe)
+  const Comp = glamorous.div()
+  const wrapper = mount(<Comp />, {context})
+  wrapper.unmount()
+  expect(unsubscribe).toHaveBeenCalled()
+})
+
+test('ignores context if a theme props is passed', () => {
```
kentcdodds: Also an awesome test :+1:

src/__tests__/theme-provider.test.js
```diff
@@ -0,0 +1,108 @@
+/* eslint func-style:0 */
+import React, {Component} from 'react'
+import {render, mount} from 'enzyme'
+import * as jestGlamorReact from 'jest-glamor-react'
+import toJson from 'enzyme-to-json'
+// eslint-disable-next-line import/default
```
kentcdodds: Hmmm... Why was this necessary?

src/__tests__/theme-provider.test.js
```diff
@@ -0,0 +1,108 @@
...
+    ),
+  ).toMatchSnapshotWithGlamor()
+})
+
+test('theme properties updates get propagated down the tree', () => {
+  class Parent extends Component {
+    state = {
+      padding: '10px',
+    }
+    setPadding = () => this.setState({padding: '20px'})
```
kentcdodds: nit-pick, probably don't need this to be a public class field because you're calling it on the instance :) could probably just be a normal method.

Also, I think that enzyme exposes [a mechanism](https://github.com/airbnb/enzyme/blob/master/docs/api/ReactWrapper/setState.md) to set state on an instance, so maybe we should just do it that way instead?

src/__tests__/theme-provider.test.js
```diff
@@ -0,0 +1,108 @@
...
+  const Child = glamorous.div(
+    {
+      color: 'red',
+    },
+    (props, theme) => ({padding: theme.padding}),
+  )
+  const wrapper = mount(<Parent />)
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
+  wrapper.instance().setPadding()
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
```
kentcdodds: Same notes on this one about the snapshot name and the fact you shouldn't need `toJson` :)

src/__tests__/theme-provider.test.js
```diff
@@ -0,0 +1,108 @@
...
+})
+
+test('merges nested themes', () => {
```
kentcdodds: Great test. Could we add also the fact that they are truly merged to this test by making the first theme include a `margin: 1`, so we can see that the `margin: 2` from the second theme is what is actually rendered.

Also, note that if your unit is `px` you can make it a number instead of a string. So `{padding: '1px'}` can be: `{padding: 1}` :)

src/__tests__/theme-provider.test.js
```diff
@@ -0,0 +1,108 @@
...
+  expect(
+    mount(
+      <ThemeProvider theme={{padding: '1px'}}>
+        {false && <div />}
+      </ThemeProvider>,
+    ),
+  ).toMatchSnapshotWithGlamor()
+})
+
+test('cleans up outer theme subscription when unmounts', () => {
```
kentcdodds: Boo yeah!

src/__tests__/theme-provider.test.js
```diff
@@ -0,0 +1,108 @@
...
+  wrapper.unmount()
+  expect(unsubscribe).toHaveBeenCalled()
+})
+
+test('does nothing when receive same theme via props', () => {
+  const theme = {margin: '2px'}
+  const wrapper = mount(<ThemeProvider theme={theme} />)
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
+  wrapper.setProps({theme})
+  expect(toJson(wrapper)).toMatchSnapshotWithGlamor(`enzyme.mount`)
```
kentcdodds: Same notes on the snapshot name. :+1:

src/theme-provider.js
```diff
@@ -0,0 +1,75 @@
+import React, {PropTypes, Component} from 'react'
```
kentcdodds: Could you use the new `react-compat.js` for `PropTypes`? See `src/index.js`

src/index.js
```diff
@@ -221,3 +256,4 @@ function joinClasses(...classes) {
 //   globalObject.glamorous = glamorous
 // }
 export default glamorous
+export {ThemeProvider}
```
kentcdodds: You can feel free to delete the commented out code above this. That was a mistake.

Also, as discussed, I think that if you add: `Object.assign(glamorous, {ThemeProvider})` that should make our export be what we want.

src/theme-provider.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * This is a component which will provide a theme to the entire tree
+ * via context and event listener (because pure components block context updates)
+ * @param {Object} theme the theme object..
+ */
+class ThemeProvider extends Component {
+  broadcast = brcast(this.props.theme)
+
+  // create theme, by merging with outer theme, if present
+  getTheme = passedTheme => {
```
kentcdodds: This is always called on the instance `this.getTheme` so it doesn't need to be a public class field and can be a normal method.

src/theme-provider.js
```diff
@@ -0,0 +1,75 @@
...
+      this.broadcast.setState(this.getTheme(nextProps.theme))
+    }
+  }
+
+  componentWillUnmount() {
+    this.unsubscribe && this.unsubscribe()
+  }
+
+  render() {
+    return this.props.children ?
```
kentcdodds: Perfect :+1:

kentcdodds:
Oh, also, could you [add yourself](https://github.com/paypal/glamorous/blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) (and those who've so kindly reviewed this PR) to [the contributor's table](https://github.com/paypal/glamorous/blob/master/README.md#contributors)?

Thanks so much everyone! This has been super cool :)

src/theme-provider.js
```diff
@@ -0,0 +1,75 @@
+import React, {PropTypes, Component} from 'react'
+import brcast from 'brcast'
+
+export const CHANNEL = '__glamorous__'
+
+/**
+ * This is a component which will provide a theme to the entire tree
+ * via context and event listener (because pure components block context updates)
```
kentcdodds: The build is failing because this line is too long. You might run `npm start validate` before you commit to check that :+1:

kwelch:
@kentcdodds if there is a plan to export `__CHANNEL__` for use by a user component then I would absolutely recommend a decorator. This would match the `withRouter` pattern in react-router. 

Might not be a requirement of this PR, but I can add a PR that add it once this is completed.

vesparny:
@kentcdodds I would not expose `CHANNEL`.
I think it's better to create a `withTheme` HoC people can use to wrap non glamorous components.

@kwelch 
We could use recompose and make it a one-liner 
https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withContext.js

src/__tests__/index.js
vesparny: I guess because it is not possible to use ES6 classes with `fest.fn`

src/__tests__/theme-provider.test.js
vesparny: I think I copied/pasted from the other test file.
I will remove this line from both files as it is not necessary

kwelch:
That should cover my concerns.  👍

rollup.config.js
vesparny: I've tried your solution and it works only by using `named` in the rollup.config.

Using `default` does not work if we export more than just `default`
BTW if we use `named` I believe we do not even need to `Object.assign(glamorous, {ThemeProvider})`

My rollup knowledge is too limited tho

kentcdodds: We may have to utilize that commented out code to get what we want. I've added a `dist-test` to make sure we don't break the way we currently export the module and when I changed the `exports` in the rollup config to `named` it breaks the `require` API (you have to do `require('glamorous').default` which I really want to avoid).

That commented out code [was an accident](https://medium.com/@kentcdodds/please-don-t-commit-commented-out-code-53d0b5b26d5f), but it came from [match-sorter](https://github.com/kentcdodds/match-sorter) to overcome limitations noted in [the rollup docs for `exports`](https://github.com/rollup/rollup/wiki/JavaScript-API#exports).

I'll probably merge the `dist-tests` here in a minute. If you could rebase onto that and try to make it work that'd be great. If you need a hand, let me know!

src/index.js
```diff
@@ -204,20 +246,5 @@ function joinClasses(...classes) {
   return classes.filter(Boolean).join(' ')
 }
 
-// some manual magic umd here because Rollup isn't capable of
```
kentcdodds: Thanks for removing this. But I think we might need it (as mentioned earlier). 😅

rollup.config.js
vesparny: @kentcdodds I pushed a fix for that.
It feels a bit hacky. I guess we can probably fix that at build time directly via rollup. Not sure how tho.

kentcdodds: Hmmm... you know what, I feel sorry about this, but I've changed my mind. It's probably better if we don't do that weird hacky stuff and just stick with what you had before. We'll push a breaking change. Sorry about that. Do you mind going back to what you had before?

vesparny: haha :)
I agree with you.
I would not bother with `glamorous.default`

rolling back in few seconds

vesparny: done, but now the dist-test will complain

vesparny: do you want me to make it return true, then we fix it in another PR?

kentcdodds:
I'm updating this PR now. I think we're going to get the best of both worlds :)

kentcdodds:
Alrighty, I think what I did in https://github.com/paypal/glamorous/pull/24/commits/bd94a1105f9e5b2ff5dcadcc0f1746c1f6a35921 will work really well for what we're trying to do as far as exports goes. It's much less hacky and the ES export is not negatively impacted at all so that's great :)

I think this is ready to be merged. Anyone wanna take a final look at mark it go/no-go?

dist-test/index.js
```diff
@@ -56,21 +56,13 @@ function isGlamorousFunction(thing) {
     )
     return false
   }
-  if (thing.name !== 'glamorous') {
```
kentcdodds: I found that the name doesn't really matter all that much and the UMD/CJS builds have glamorous with a different name. NBD :)

kwelch:
Only thing I found was a inconsistency with the documentation on how to import `ThemeProvider`. 

Other than that from what I can tell this looks fantastic.

vesparny:
@kentcdodds should we credit styled-components somehow.
I've taken inspiration from there for the merging strategy, even if the impl is quite different

kentcdodds:
Oh yes! If anything was inspired, let's document that somehow 👍

vesparny:
Ok. How?
I've no access to my laptop now

kentcdodds:
I'd probably just put a comment at the top of `theme-provider.js`. Thanks!

vesparny:
@kentcdodds done

README.md
```diff
@@ -282,12 +285,63 @@ and `A` components which make use of `src` and `href` props).
...
+
+`glamorous` fully supports theming using a special `<ThemeProvider>` component.
+
+It provides the `theme` to all glamorous components down the tree.
+
+```jsx
+import glamorous, {ThemeProvider} from glamorous
```
kwelch: I saw the update further up. 👍

src/theme-provider.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  getChildContext() {
+    return {
+      [CHANNEL]: this.broadcast,
+    }
+  }
+
+  setOuterTheme = theme => {
+    this.outerTheme = theme
```
kentcdodds: Yeah, we should probably initialize this above. I'll do that really quick. Wanna get this merged :)

kentcdodds:
Could I get feedback on this commit? Seems like a significant change: https://github.com/paypal/glamorous/pull/24/commits/ac5bd932bdd71742e7d8c40584b7b8354885b0d1

src/theme-provider.js
```diff
@@ -29,6 +30,7 @@ class ThemeProvider extends Component {
 
   setOuterTheme = theme => {
     this.outerTheme = theme
+    this.broadcast.setState(this.getTheme())
```
kentcdodds: Moved this from `componentWillMount`. I feel like without this here, anytime the emitter broadcasted an update, this component wouldn't re-render. But I could also be totally confused 😅

kentcdodds: @developit and @vesparny, is this correct or am I confused (or both? 🙃 )

vesparny: Uhm...it would still te-render because if the emitter broadcasted an update, the parent provider would setState, and the inner would received new props, triggering a children te-render. I know it's a bit confusing. Probably your latest commit was not necessary. BTW I think I wrote a test covering that specific case

vesparny: Excuse my typos I'm on mobile, with an Italian keyboard 😬

kentcdodds: Yeah, I'll revert this. Then I think we're ready to merge!

src/theme-provider.js
kentcdodds: I think we're good actually.

kentcdodds:
Alrighty. I think we're ready to merge!

vesparny:
💪

kentcdodds:
There goes the build and release: https://travis-ci.org/paypal/glamorous/builds/220710237

🚢

vesparny:
Awesome, will work on examples tomorrow.
Thank you all, it's been so good 💥

kentcdodds:
Here's a demo! https://codesandbox.io/s/o2yq9MkQk

kentcdodds:
@vesparny, thanks so much for your help! I've added you as a collaborator on the project. Please make sure that you review the CONTRIBUTING.md file (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

PR #26, opened 2017-04-07
refactor: use html-tag-names

### LINKED ISSUE #8: html-tags
https://github.com/sindresorhus/html-tags

tsriram:
We'd been using a local file (src/dom-elements.js) which exported an array of all HTML elements. We
now use html-tag-names module which has a more comprehensive list of HTML elements.

closes #8

**What**: Refactoring the code to use `html-tag-names` module instead of local file to get the list of HTML elements.

**Why**: `html-tag-names` has more comprehensive and up-to-date list of HTML elements.

**How**: Removed the local file `src/dom-elements.js` and added `html-tag-names` as a dependency.


src/dom-elements.js
```diff
@@ -1,145 +0,0 @@
...
-  'tr',
-  'track',
-  'u',
-  'ul',
-  'var',
-  'video',
-  'wbr',
-
-  // SVG
-  'circle',
```
paulmolluzzo: Do we want to keep these `svg`-related tags in addition to the new `html-tags-names` we get from the package? In particular, `line`, `path`, `circle`, `rect`, `text`, etc. are display/shape/rendered elements in `svg` and I could imagine a scenario where I'd want a way to make a component based on these elements.

I'd prefer to keep them, even if it means that we need to `concat` the the `html-tag-names` with a hand curated list.

paulmolluzzo:
@tsriram Thanks for working on this! 🙌

src/dom-elements.js
paulmolluzzo: Also, it might not be a big deal, but if we (plan to) remove an element from the list that someone is using we should give a deprecation notice. It's possible these elements aren't being used by anyone, but how can we know for sure?

kentcdodds: Originally I had said that these would be out of scope. I didn't remember that they were actually already included. I feel like it would be better to not make this a breaking change.

So, with that said, could we also include https://github.com/wooorm/svg-tag-names and concat `htmlTagNames` and `svgTagNames` together into `domElements` in `src/index.js`?

tsriram: Cool. Have added `svg-tag-names` as well. Wonder there should be any test related to this? Actually I can't think of any test for this, but just wanted to confirm!

tsriram:
Btw, did my name break the layout of Contributors table? :grin: 

Take a look here - https://github.com/tsriram/glamorous/tree/html-tag-names

paulmolluzzo:
> Btw, did my name break the layout of Contributors table? 😁

👍  It makes you look taller than everyone else.

src/dom-elements.js
paulmolluzzo: We can test that it's an array and has a length greater than `0`. I don't know if we need to though.

src/index.js
```diff
@@ -5,7 +5,11 @@
 import React from 'react'
 import {css, styleSheet} from 'glamor'
 import shouldForwardProperty from './should-forward-property'
-import domElements from './dom-elements'
+
+const htmlTagNames = require('html-tag-names')
+const svgTagNames = require('svg-tag-names')
```
kentcdodds: Instead of `require` I'd prefer to use ESModules with `import` statements:

```javascript
import htmlTagNames from 'html-tag-names'
import svgTagNames from 'svg-tag-names'
```

src/index.js
```diff
@@ -7,6 +7,9 @@ import {css, styleSheet} from 'glamor'
 import shouldForwardProperty from './should-forward-property'
 
 const htmlTagNames = require('html-tag-names')
+const svgTagNames = require('svg-tag-names')
+
+const domElements = htmlTagNames.concat(svgTagNames)
```
paulmolluzzo: Do we need to deduplicate this array? `svg` and `iframe` are repeated, but there may be others.

kentcdodds: Good point. May be a good idea for us to dedupe...

kentcdodds: We could actually do that dedupe in the `Object.assign` calls below:

```javascript
Object.assign(
  glamorous,
  domElements.reduce(
    (getters, tag) => {
      if (getters[tag]) { // <-- this prevents overwriting
        return getters
      }
      getters[tag] = glamorous(tag)
      return getters
    },
    {},
  ),
)

// ...

Object.assign(
  glamorous,
  domElements.reduce(
    (comps, tag) => {
      const capitalTag = capitalize(tag)
      if (comps[capitalTag]) { // <-- this prevents overwriting
        return comps
      }
      comps[capitalTag] = glamorous[tag]()
      comps[capitalTag].displayName = `glamorous.${capitalTag}`
      comps[capitalTag].propsAreCssOverrides = true
      return comps
    },
    {},
  ),
)
```

Alternatively we could just do:

```javascript
domElements = htmlTagNames.concat(svgTagNames).filter((tag, index, array) => array.indexOf(tag) === index)
```

I think that should work :)

kentcdodds:
> Btw, did my name break the layout of Contributors table? 😁

Haha, yeah, don't worry about that. We can only do so much there :) The more contributors we add the less you'll stick out like that :) Thanks again for working on this!

src/index.js
kentcdodds: To be clear, let's do that second option with the `filter` rather than messing with the reducer functions.

src/index.js
tsriram: Yup, I'd prefer import too. But it doesn't work when I do `import htmlTagNames from 'html-tag-names'`. Is it because the modules (both `html-tag-names` & `svg-tag-names` doesn't have a default export? In package.json, `main` is pointing to a JSON file which has the array of tag names.

kentcdodds: Ah, perhaps. That's fine. No worries :)

kentcdodds:
Thank you so much for this!

tsriram:
Awesome! Thank you @kentcdodds and @paulmolluzzo :smile:

PR #27, opened 2017-04-07
Examples - Add an example to test glamorous with enzyme and jest

### LINKED ISSUE #21: Add example using jest-glamor-react
https://github.com/kentcdodds/jest-glamor-react
Because it's super cool!

pksjce:
Add an example to test glamorous with enzyme and jest

It would easier for the community to adopt glamorous by experiencing its testability easily.

I picked up the example in the docs of glamor-jest-react and converted the same to this use case

Fixes #21


pksjce:
Not sure why the build is failing! https://travis-ci.org/paypal/glamorous/builds/219600857
`jest` seems to be running the dist folder -\o/-
Help?

paulmolluzzo:
> Not sure why the build is failing! https://travis-ci.org/paypal/glamorous/builds/219600857
> jest seems to be running the dist folder -\o/-
> Help?

I think we need to add the following to our `package.json` to be sure we're only testing the `src` directory:

```js
"collectCoverageFrom": ["src/**/*.js"],
```

@kentcdodds Can you confirm this please? I tested locally and the output matches from `master` and this PR when I add this to `package.json`, but I'm not familiar with `jest` enough to know for sure.

kentcdodds:
That sounds correct to me @paulmolluzzo 👍

examples/glamorous-jest-react/index.js
```diff
@@ -1,19 +1,20 @@
 import React from 'react'
-import * as glamorous from 'glamorous'
+/* eslint-disable */
+import * as glamorous from 'glamorous';
```
paulmolluzzo: Shouldn't this be `import glamorous from 'glamorous'`? `glamorous` only has one export.

examples/glamorous-jest-react/README.md
```diff
@@ -0,0 +1,9 @@
+## Example for `glamorous` with `jest` and `enzyme`
```
kentcdodds: Could you rename this directory to simply `examples/with-jest`. I think this is going to be our standard way to recommend people use this library with `jest` in general :+1:

examples/glamorous-jest-react/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+### Steps to run
```
kentcdodds: Could you add `git clone` instructions as well. Want to make this as beginner-friendly as possible :)

examples/glamorous-jest-react/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,67 @@
...
+  <h1
+    class="css-1tnuino"
+  >
+    Hello World, this is my first glamor styled component!
+  </h1>
+</section>
+`;
+
+exports[`enzyme.shallow 1`] = `
+<glamorous(section)>
```
kentcdodds: Hmmm... I'm not sure why we're not getting the CSS output for the shallow render 🤔 could you look into it a little?

examples/glamorous-jest-react/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+/* eslint-disable */
```
kentcdodds: Why is this disabled? I'd prefer to keep this file linted.

examples/glamorous-jest-react/index.js
```diff
@@ -0,0 +1,20 @@
...
+import * as glamorous from 'glamorous';
```
kentcdodds: `glamorous` exports itself as the `default` export, so this should be `import glamorous from 'glamorous'`

examples/glamorous-jest-react/index.test.js
```diff
@@ -0,0 +1,22 @@
+import {matcher, serializer} from 'jest-glamor-react'
+import React from 'react'
+import * as enzyme from 'enzyme'
+import toJson from 'enzyme-to-json'
```
kentcdodds: Instead of using `toJson` everywhere, we could actually configure Jest to use the serializer like [this](https://github.com/adriantoine/enzyme-to-json#serializer) :)

examples/glamorous-jest-react/index.test.js
```diff
@@ -0,0 +1,22 @@
...
+import {Wrapper, Title} from './index'
+
+expect.addSnapshotSerializer(serializer)
```
kentcdodds: Could you add a comment above this line that says:

```
// add the serializer and `toMatchSnapshotWithGlamor` matcher
// you can do this in a setup file instead so you don't have to do it
// for all of your tests
// https://facebook.github.io/jest/docs/configuration.html#setupfiles-array
```

Either that or maybe we should just use a setup file. Whichever you prefer :)

pksjce:
@paulmolluzzo - Thanks for the tip. That fixed my build!

examples/glamorous-jest-react/__snapshots__/index.test.js.snap
pksjce: I thought the point of shallow render is to not get detailed snapshots. I've usually seen no depth in shallow snapshots. Are you expecting the css to show up as class in this snapshot like render/mount?

examples/glamorous-jest-react/index.test.js
pksjce: I think this comment is a good solution. If I put it in a setup file for this example, it's significance might not be easily noticeable?

paulmolluzzo:
@pksjce broken again. 😜 Looks like a lint issue?

If you aren't doing it already, you may want to run `yarn start validate` before committing/pushing. It's a faster result than waiting on Travis. I believe you already set this up, but there's also good info in [CONTRIBUTING.md](https://github.com/paypal/glamorous/blob/master/CONTRIBUTING.md).

pksjce:
@paulmolluzzo - So what happens is, if I have `node_modules` folder inside the `examples/with-jest`, this lint error does not happen. I will try pointing `glamorous` within the source code.

paulmolluzzo:
>  So what happens is, if I have `node_modules` folder inside the `examples/with-jest`, this lint error does not happen.

Travis isn't recursively installing `npm` packages. We definitely don't want `node_modules` committed, so we should either ignore the line or all examples I think.

@kentcdodds what do you think? What build tests should be run/skipped for this directory?

kentcdodds:
Hi! Thanks for all this work!

I definitely don't want the build to fail because of examples. However it would be super neat if we ran the examples during the build (after the publish step). So maybe we could have a script that installs the example's dependencies... And then runs the tests.

examples/with-jest/index.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import glamorous from 'glamorous'
+import glamorous from '../../src'
```
paulmolluzzo: I don't know about this. I would prefer that the examples are self-contained and can be ripped out and still work. This way people can `cp -r examples/with-jest/ ~/my-new-project` and build something new.

kentcdodds:
So we'd want to add the examples directory to the .eslintignore

examples/with-jest/index.js
kentcdodds: Agreed, let's make glamorous part of the dependencies for this example.

paulmolluzzo:
@pksjce yes, thank you for this! 🙏 Examples are super helpful when learning something new, so this will be very useful for people new to `glamorous` or `jest`.

pksjce:
@kentcdodds - I did add that inside the `package-scripts.js`. But it takes a looong time to install dependencies. The `validate` becomes really slow.

pksjce:
Yay! Build passed!

kentcdodds:
Yeah, I wouldn't make it part of validate, I'd add a nested script to validate called examples and that would do the installation and testing. Then I would add: `yarn start validate.examples` after [this](https://github.com/paypal/glamorous/blob/master/.travis.yml#L14).

pksjce:
Oh Cool! I think I could do that. Let me try.

pksjce:
```
    examples: {
      withJest: {
        script: series(
          'cd with-jest',
          'npm install',
          'npm run test',
          'cd ../'
        ),
      },
    },
    validate: {
      description: 'This runs several scripts to make sure things look good before committing or on clean install',
      default: concurrent.nps('lint', 'build', 'test'),
      examples: {
        script: series(
        'cd examples',
        'nps examples.withJest',
        'cd ../'
        ),
      },
    },
```

Something like this? I don't understand nps very well. Any utility function I'm missing to move up and down the directories?

kentcdodds:
That looks great! Probably don't need the cds on the validate.examples script (and then you'd need to update the cds on the examples.withJest). Other than that it looks great!

.all-contributorsrc
```diff
@@ -64,7 +64,7 @@
         "doc",
         "example"
       ]
-    },
+    },`
```
kentcdodds: This appears to be a mistake.

examples/with-jest/index.test.js
```diff
@@ -0,0 +1,19 @@
+import {matcher, serializer} from 'jest-glamor-react'
+import React from 'react'
+import {shallow, render, mount} from 'enzyme'
+import {Wrapper, Title} from './index'
+
+expect.addSnapshotSerializer(serializer)
```
kentcdodds: Let's not forget the comment here. Or actually, I think it would be better to do the setup file because that's what people should do in real applications.

package-scripts.js
```diff
@@ -55,9 +55,24 @@ module.exports = {
         'semantic-release post'
       ),
     },
+    examples: {
+      withJest: {
+        description: 'This jumpstarts and validates the with-jest example.',
+        script: series(
+          'cd examples/with-jest',
+          'npm install',
```
kentcdodds: Let's switch this to yarn 👍

package.json
```diff
@@ -95,7 +95,8 @@
     },
     "snapshotSerializers": [
       "enzyme-to-json/serializer"
-    ]
+    ],
+    "collectCoverageFrom": ["src/**/*.js"]
```
kentcdodds: Maybe we should add [roots](https://facebook.github.io/jest/docs/configuration.html#roots-array-string) config so it only looks for tests in the `src` directory.

examples/with-jest/index.test.js
pksjce: How do I get `expect` inside setupFile? 
`ReferenceError: expect is not defined` is what I get.

kentcdodds: Ah, actually, we should probably use: [`setupTestFrameworkScriptFile`](https://facebook.github.io/jest/docs/configuration.html#setuptestframeworkscriptfile-string)

pksjce:
Hey @kentcdodds - I think I was able to address all the review comments :)

.travis.yml
```diff
@@ -9,6 +9,7 @@ node_js:
   - '6'
 script:
   - yarn start validate
+  - yarn start validate.examples
```
kentcdodds: Awesome. If you could move this to `after_success` (after the `yarn start release` script), that would be great. That way the build doesn't break if the examples are broken. I just don't want a broken example to prevent someone's contribution making it into the lib or stopping a release. Thanks!

README.md
```diff
@@ -14,7 +14,7 @@ small footprint (<5kb gzipped), and great performance (via [`glamor`][glamor]).
 [![downloads][downloads-badge]][npm-stat]
 [![MIT License][license-badge]][LICENSE]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-7-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-0-orange.svg?style=flat-square)](#contributors)
```
kentcdodds: I think this happened because there was an issue in `.all-contributorsrc` earlier. Now that you've fixed it, could you run: `npm start contrib.gen`? Thanks!

README.md
```diff
@@ -474,7 +472,7 @@ MIT
 [roadmap-badge]: https://img.shields.io/badge/%F0%9F%93%94-roadmap-CD9523.svg?style=flat-square
 [roadmap]: https://github.com/paypal/glamorous/blob/master/other/ROADMAP.md
 [examples-badge]: https://img.shields.io/badge/%F0%9F%92%A1-examples-8C8E93.svg?style=flat-square
-[examples]: https://github.com/paypal/glamorous/blob/master/other/EXAMPLES.md
+[examples]: https://github.com/paypal/glamorous/blob/master/examples/README.md
```
kentcdodds: On second thought, instead of linking to the README, let's link to the directory itself (and github will render the README, but people will also see the other folders).

So:

```markdown
[examples]: https://github.com/paypal/glamorous/blob/master/examples
```

examples/with-jest/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,71 @@
...
+  <h1
+    class="css-1tnuino"
+  >
+    Hello World, this is my first glamor styled component!
+  </h1>
+</section>
+`;
+
+exports[`enzyme.shallow 1`] = `
+.css-1otybxc,
```
kentcdodds: Glad you got this to work :+1:

examples/with-jest/testSetup.js
```diff
@@ -0,0 +1,4 @@
+import {matcher, serializer} from 'jest-glamor-react'
+
+expect.addSnapshotSerializer(serializer)
```
kentcdodds: Perfect. Could you add a comment above this line that says:

> // This is what adds the CSS to the output snapshot

Thanks

examples/with-jest/testSetup.js
```diff
@@ -0,0 +1,4 @@
...
+expect.extend(matcher)
```
kentcdodds: Could you add a comment above this line that says:

> // this adds `toMatchSnapshotWithGlamor` to `expect` and makes the snapshot diff output look nice in the terminal

I realize that this may be more in the scope of the `jest-glamor-react` docs, but I think this could still be helpful :) Thanks!

package.json
```diff
@@ -95,7 +95,8 @@
     },
     "snapshotSerializers": [
       "enzyme-to-json/serializer"
-    ]
+    ],
+    "roots": ["src"]
```
kentcdodds: Could you format `"src"` on a new line? Next time yarn/npm does something to this file, it'll reformat it and I'd prefer it to not change this line :)

pksjce:
Awesome!

paulmolluzzo:
This is awesome! Excellent work @pksjce! 👏👏

@kentcdodds merging this release v3.0.0 because the commits said "BREAKING CHANGE: none". This shouldn't even really bump the version of `glamorous` according to semver (I think). Should we better document this in CONTRIBUTING.md?

kentcdodds:
Whoops! 😅 Yeah, I've been thinking about completely removing commitizen and instead have better docs on the message format. In any case, improved docs are in order.

Also, a bot that labels the pull request based on the commit message format would be awesome... Something to think about.

PR #30, opened 2017-04-07
chore(config): Add union merge strategy for README.md

paulmolluzzo:
**What**:
With this inclusion git will attempt to keep multiple changes to a README to avoid conflicts

**Why**:
This is something I learned about [here](https://github.com/sindresorhus/refined-github/pull/23). That project had lots of changes to the README, and in `glamorous` the `all-contributors` might cause unnecessary conflicts. This could be nice to have and avoid merging/rebasing a PR.

**How**:
Small change in the `.gitattributes` file.

More info is available here: https://github.com/isaacs/github/issues/487

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 * text=auto
 *.js text eol=lf
+README.md merge=union


paulmolluzzo:
I'm wondering if we should also add `.all-contributorsrc` as well. It's another file that will get mostly additions where we'll want to keep both. Look at the conflicting files for #27 as an example.

kentcdodds:
Yeah, I still don't really understand what this does 😅

paulmolluzzo:
If there are two additions on the same line it'll keep both additions.

kentcdodds:
Ah, yeah, better not then...

PR #31, opened 2017-04-07
fix(umd): add json plugin

kentcdodds:
**What**: This adds `rollup-plugin-json`

**Why**: Because we broke the rollup build when we started requiring packages that use `main` as a `json` file 😄 

**How**: add the plugin, also refactor things a little bit to be cleaner.

cc: @tsriram

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "rollup": "^0.41.6",
     "rollup-plugin-babel": "^2.7.1",
     "rollup-plugin-commonjs": "^8.0.2",
+    "rollup-plugin-json": "^2.1.0",
     "rollup-plugin-node-resolve": "^3.0.0",
     "rollup-plugin-uglify": "^1.0.1",
     "semantic-release": "^6.3.6",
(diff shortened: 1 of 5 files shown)


kentcdodds:
I added a PR to fix the dependency ci issue: https://github.com/rollup/rollup-plugin-json/pull/25

kentcdodds:
Merging this as-is considering it's breaking anyone using the UMD build via unpkg.com (like my intro blogpost :wink:)

Feel free to still review and we can address any needed changes.

paulmolluzzo:
I'll review tonight.

> Merging this as-is considering it's breaking anyone using the UMD build via unpkg.com (like my intro blogpost 😉)

🙈

tsriram:
So `npm start validate` is supposed to fail without this change?

kentcdodds:
No, `npm start validate` wont fail based on the UMD build. Though we could make it do that like I do in `match-sorter` with the [`dist-test`](https://github.com/kentcdodds/match-sorter/blob/master/dist-test/index.js). Not a bad idea actually...

Issue #10, opened 2017-04-05
Improve documentation


kentcdodds:
Could I get some people to look through the docs and file issues or [submit PRs](http://makeapullrequest.com) where it's unclear or could use an example? Thanks!

paulmolluzzo:
There could definitely be some additional info included for SSR I think. See https://github.com/zeit/next.js/pull/1620#issuecomment-291747882

Also, there are some great examples included in [this Medium post](https://medium.com/@kentcdodds/introducing-glamorous-fb3c9f4ed20e) that might fit nicely in the README.

kentcdodds:
Great idea! Would you like to make a pull request?

paulmolluzzo:
@kentcdodds you bet!

paulmolluzzo:
The README doesn't have examples o led extending components with `glamorous`, which is very accessible and powerful with this package.

If someone wants to add examples I already wrote a few here: https://github.com/paulmolluzzo/glamorous-jsxstyle/blob/master/README.md#usage

Can certainly be brought over here, and is a good task for a beginner to GitHub and/or `glamorous`. 😉

paulmolluzzo:
@kentcdodds we should have a 'good for beginner' label (or something equivalent).

dantesolis:
Hey guys I just read the medium blog post on glamour. Is anyone working on this and if not can I take it? 

Thanks

paulmolluzzo:
@dantesolis go for it! 😀

This is more of an "open task" for now, so no one person needs to claim it. If someone sees an area that can be improved then a PR is welcome.

dantesolis:
@paulmolluzzo thanks are there any other additional task that a beginner could tackled which have  not yet been claimed? Thanks

paulmolluzzo:
This is the only issue with the `good first issue` label right now. 😕 We do plan to use that label so keep an eye out for it.

It's possible that you'll find an issue yourself using `glamorous` in some project, and finding new issues is also very helpful.

kentcdodds:
This is an ongoing effort. I'll go ahead and close this now.

PR #33, opened 2017-04-09
add simple example in CodeSandbox

UsulPro:
**What**:
Added a link to the WebpackBin with simle example to README.MD

**Why**:
It helps people to start playing with it

**How**:
Creating a new [Bin](https://www.webpackbin.com/bins/-KhHc0veo9yP0wgtvyko) with `glamorous` example

In order to help with #10


paulmolluzzo:
@UsulPro thanks for the PR! Please add yourself to the contributors list. You may want to review the info in [CONTRIBUTING.md](https://github.com/paypal/glamorous/blob/master/CONTRIBUTING.md) to find out how.

kentcdodds:
Cool! Thanks! I haven't been able to look at it yet because I'm on mobile and webpack bin doesn't support mobile. But I'm thinking that I'd rather use https://codesandbox.io anyway because that's what http://kcd.im/glamorous-help uses and I think I'd rather encourage people to learn and use one tool for something like this. Thanks so much!

kentcdodds:
Just saw the example. It's a great start. If we could move it to https://codesandbox.io that would be great. Also, I'd love to have it pre-loaded with a few things (and comments) to demonstrate some of the features so people can start playing with things right away (without having to reference the docs as much). What do you think?

Thanks again! This is great :+1:

pksjce:
How about adding this whole setup as a code example and then linking to the bin as a demo?

UsulPro:
@kentcdodds Thank you for pointing me on the `codesandbox.io`! I did not know about it before 👍 

Do you like this two examples:

[simple example](https://codesandbox.io/embed/pk744ggp?module=D8wny)
[Glamorous Component](https://codesandbox.io/embed/EJ4kmOV4?module=D8wny)

My idea was to provide a link after each example in readme to play with it. But the last link confuses me because I can't make it working via pointing to the specified file (`app.js`). At the same time [this sandbox](https://codesandbox.io/s/EJ4kmOV4) in general works fine. So maybe it's only my *Chrome/Windows* issue or it'd be better to provide another kind of link?

kentcdodds:
> My idea was to provide a link after each example in readme to play with it.

I love that idea! Let's do that.

If you want to include the glamorous logo, rather than linking to it with an Img tag, maybe you could just render it like we do in http://kcd.im/glamorous-help ?

I'm not sure about that last link. It wasn't working for me either 🤔

kentcdodds:
@pksjce could you explain what you mean?

CompuIves:
I'm the CodeSandbox creator quickly jumping in to clarify why the links didn't work :).

There is a small confusion with the embed view, in embed view the previewer is automatically set to 'module view', which means it only renders the module that's opened in the code editor. 
Since in the links 'app.js' is selected as default module it only previewed 'app.js' which in the second case is a gray box without content.

The solution is to change this:
https://codesandbox.io/embed/EJ4kmOV4?module=D8wny
to this:
https://codesandbox.io/embed/EJ4kmOV4

I completely understand the confusion, I've added some extra explanation to codesandbox.io to fix this.

UsulPro:
@CompuIves Thank you for your response! Just to clarify: will it start to work if I provide default values for props?

@kentcdodds I'm impressed with how you generate a logo from emoji :) 
In this case how do you think is it Okey if the code in codesandbox.io will differ from the readme example?

CompuIves:
@UsulPro Yes, then it should work! I'm thinking about giving the user the ability to specify props in the preview so they can play with exported components, but for now it's either calling render in `app.js` or using default props.

README.md
```diff
@@ -86,6 +86,8 @@ function App() {
 }
 ```
 
+> Try this out in your browser [here](https://codesandbox.io/s/pk744ggp)!
```
kentcdodds: Looks great. I don't mind that it differs slightly. I would prefer to keep everything in a single file rather than have multiple files. That way people don't get confused.

README.md
```diff
@@ -270,6 +272,8 @@ function MyUserInterface({name, tagline, imageUrl, homepage, size}) {
 }
 ```
 
+> Try this out in your browser [here](https://codesandbox.io/s/EJ4kmOV4)!
```
kentcdodds: I think the glamorous logo looks kinda funny in the circle there. Why don't you find a funny cat picture or something to use instead ;-) And the tagline can be anything else :)

UsulPro:
@kentcdodds Done!

[Simple example](https://codesandbox.io/s/pk744ggp)

[Glamorous Component](https://codesandbox.io/s/EJ4kmOV4)

kentcdodds:
Hi @UsulPro! Hmmm, the diff in this PR makes it look like the only thing that changed is the contributors table in the README... Could you look into that? Also, could you make both examples use the `glamorous.div` API as well? I'd like to highlight both in what will be the first examples for many people :) Also, I LOVE the image you chose for the other example :heart: :octocat:

UsulPro:
Thank you @kentcdodds 😸 

Could you please clarify whether you mean `glamorous.div` or maybe `glamorous.Div`? And should other elements (`Span`, `A`, `Img`) also use this API?

Oh, I lost these links when merging with master (( sorry for that!

kentcdodds:
> Could you please clarify whether you mean glamorous.div or maybe glamorous.Div? And should other elements (Span, A, Img) also use this API?

Sure, as indicated in [the top of the README](https://github.com/paypal/glamorous#this-solution), it shows both of the main APIs you have available to you, one is the component API (`glamorous.Div`), and the other is the componentFactory API (`glamorous.div`). I'd like to highlight both of these APIs in your examples :) Thanks!

UsulPro:
@kentcdodds sorry for late response

I have corrected both examples so that they use both API variants, and added some comments:

[Simple example](https://codesandbox.io/s/pk744ggp)
[Glamorous Component](https://codesandbox.io/s/EJ4kmOV4)

Is this what you expected? Let me know please if you want me to change/fix something

kentcdodds:
Cool! That's close, I've just forked them and made a few changes:

[Simple example](https://codesandbox.io/s/mDLZ1oKn)
[Glamorous Component](https://codesandbox.io/s/wmRo8OKDm)

What do you think?

UsulPro:
@kentcdodds Great!
Do you want to use these forks for examples? It seems to me that it'll be more convenient if you need to correct something in the future

kentcdodds:
That sounds fine, thank you!

kentcdodds:
Looks like there's a merge conflict in this PR. [This may help](https://egghead.io/lessons/javascript-how-to-rebase-a-git-pull-request-branch) :)

UsulPro:
Thank you for this short guide 👍 
I've squashed it as well

I noticed that `all-contributors-cli` is already installed as a devDependency. So maybe a few words about how to use it could help to do it :)

kentcdodds:
Thank you!

PR #38, opened 2017-04-11
docs(examples): Added link to example created within css-in-js repo

kwelch:
**What**:
Added link to example created within the css-in-js repository.

**Why**:
Examples = great documentation. [Oh and Kent asked me to](https://github.com/MicheleBertoli/css-in-js/pull/52#issuecomment-292645192)

**How**:
Add link and small preview description to examples markdown. 

I wasn't sure what it should look like or if it need text. I am open to recommended changes.


kwelch:
That was easier than I expected.

paulmolluzzo:
Hmm, shouldn't we link to the blob and not `master`? 🤔 Current link is the `HEAD` of `master`, which might eventually break.

Blob link: https://github.com/MicheleBertoli/css-in-js/blob/890e0de6b6324d6b03be03f942f7969be2950fe2/glamorous/button.js

kentcdodds:
True, but we also want it to point to the latest version in case it's updated. I think we'll just hope that it doesn't break.

Issue #43, opened 2017-04-11
Make glamorous faster


kentcdodds:
We haven't done any benchmarking, but @garetmckinley [is going to](https://twitter.com/garetmckinley/status/851689363050254336):

> I'm going to benchmark glamorous against styled-components this week, so we will see what the numbers say 😊 dont worry, I still love 💅

So let's find bottlenecks and make them faster. There's definitely some low-hanging fruit from my original implementation that didn't need to be that way. Let's start in [the `render`](https://github.com/paypal/glamorous/blob/master/src/index.js#L73-L98) function which is probably the hottest code path of the codebase.

Anyone wanna try setting up some benchmarking to determine our bottlenecks and start speeding things up? Maybe @developit could give us some pointers...

kentcdodds:
I think that we could maybe write a babel plugin to take this:

```javascript
const MyDiv = glamorous.div({padding: 20}, ({big}) => ({fontSize: big ? 20 : 30}))
```

to this:

```javascript
const MyDiv = glamorous.div('css-dfk22d34', ({big}) => ({fontSize: big ? 20 : 30}))
```

Where the static styles are pre-processed and the resulting styles are extracted to a separate stylesheet, and we work with glamor's rehydration API to hook things up again. We'd need to update the `glamorousComponentFactory` API to accept a string class name, but that _should_ be trivial. In any case, this would probably speed things up significantly.

I just realized that based on the way we merge glamor class names with the styles they represent into a single class name may cause issues with this, but let's start with components that have static styles only, then maybe we can get it working with a combination of static and dynamic styles.

Anyone wanna write a plugin for this? If you've not made a Babel plugin before or worked with ASTs before, you can learn about it from [my talk](https://youtu.be/CFQBHy8RCpg?list=PLV5CVI1eNcJgNqzNwcs4UKrlJdhfDjshf) :)

developit:
You can skip an object copy by removing [this line](https://github.com/paypal/glamorous/blob/master/src/index.js#L75).  Instead of "plucking" className off of props, simply use `props.className`.  In order to have your custom value take precedence over [`props.className` here](https://github.com/paypal/glamorous/blob/master/src/index.js#L95-L98) just place the inline property after the rest spread.

Other things: remove `map()`.

```js
      render() {
        const props = this.props
        const {toForward, cssOverrides} = splitProps(props, GlamorousComponent)
        const theme = Object.assign({}, this.state.theme)
        // create className to apply
[...]

PR #44, opened 2017-04-11
feat(perf): exit early when filtering props

kentcdodds:
**What**: This adds an `if` statement to exit early in the `splitProps` function

**Why**: If the rootEl is a component, then we can forward all props to it without worrying about attaching props to DOM nodes. So we can skip the reduce function entirely.

**How**: Running the tests in watch mode and hitting random keys until they passed.

Helps with #43


kentcdodds:
Unfortunately, we currently have no way to measure whether this even makes an impact, but this is something that's been on my mind. I'm pretty sure it's a bit of a bottleneck so it's probably a good idea :)

Issue #41, opened 2017-04-11
withTheme() HoC for theming normal components


vesparny:
As discussed with @kwelch in #24 we might want to provide a `withTheme()` HoC that allows using theming capabilities with non-glamorous components.

```jsx
import {withTheme} from 'glamorous'

const Hello = ({theme, ...rest}) =>
  <div style={{color: theme.colors.primary}} {...rest} /> 

export default withTheme(Hello)
```

`withTheme()` will just receive the context, setup subscription and return the wrapped component with a `theme` prop.

Am I correct @kwelch ?

kentcdodds:
Looks good to me!

kwelch:
Yes, this should isolate context to within the library so users don't have to use an experimental API (knowingly).

vesparny:
Is anyone willing to take this one? Otherwise I can do it.

kentcdodds:
Let's wait a day or so to see whether anyone else wants to do it. Unless you need it now if course. The more the merrier!

kkemple:
I would love to jump on this, been working with HOCs a lot lately.

kentcdodds:
Awesome! 👍

kkemple:
@kentcdodds sent a PR, mostly there but there are some questions that need answering before its ready to go! :smile:

PR #45, opened 2017-04-11
feat(HOC): add `withTheme` HOC

kkemple:
**What**:
Added a HOC (higher order component) called `withTheme` that exposes theme context as a prop to passed in component.

**Why**:
It was requested in Issue #41 

**How**:
I first created a new file to contain the HOC, I then added to global exports. I also added tests.

There are a few comments in there as some things are unclear still. Mainly, what do we want to do when there is no context found (no ThemeProvider further up the tree). There is also some concern on naming of things. Not sure what you want the exported HOC class name to be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,72 @@ const Title = glamorous.h1({
 </ThemeProvider>
 ```
 
+`glamorous` also exports a `withTheme` higher order component (HOC) so you can access your theme in any component!
+> Try this out in your browser [here](https://codesandbox.io/s/qYmJjE4jy)!
+
+```jsx
+import glamorous, {ThemeProvider,  withTheme} from glamorous
+
+// our main theme object
+const theme = {
+  main: {color: 'red'}
+}
+
+// a themed <Title> component
+const Title = glamorous.h1({
+  fontSize: '10px'
+}, (props, theme) => ({
+  color: theme.main.color
+}))
+
+// normal component that takes a theme prop
+const SubTitle = ({children, theme: {color}}) => (
+  <h3 style={{color}}>{children}</h3>
+);
+
+// extended component with theme prop
+const ThemedSubTitle = withTheme(SubTitle);
+
+<ThemeProvider theme={theme}>
+  <Title>Hello!</Title>
+  <ThemedSubTitle>from withTheme!</ThemedSubTitle>
+</ThemeProvider>
+```
+
+Or if you prefer decorator syntax:
+
+```jsx
+import React, {Component} from 'react';
+import glamorous, {ThemeProvider,  withTheme} from glamorous
+
+// our main theme object
+const theme = {
+  main: {color: 'red'}
+}
+
+// a themed <Title> component
+const Title = glamorous.h1({
+  fontSize: '10px'
+}, (props, theme) => ({
+  color: theme.main.color
+}))
+
+// extended component with theme prop
+@withTheme
+class SubTitle extends Component {
+  render() {
+    const {children, theme: {color}} = this.props;
+    return <h3 style={{color}}>{children}</h3>;
+  }
+}
+
+<ThemeProvider theme={theme}>
+  <Title>Hello!</Title>
+  <SubTitle>from withTheme!</SubTitle>
+</ThemeProvider>
+```
+> `withTheme` expects a `ThemeProvider` further up the render tree and will warn in `development` if one is not found!
+
 ### Server Side Rendering (SSR)
 
 Because both `glamor` and `react` support SSR, `glamorous` does too! I actually
--- a/src/__tests__/with-theme.js
+++ b/src/__tests__/with-theme.js
@@ -1,0 +1,130 @@
+/* eslint func-style:0 */
+import React, {Component} from 'react'
+import {render, mount} from 'enzyme'
+
+import withTheme from '../with-theme'
+import ThemeProvider, {CHANNEL} from '../theme-provider'
+import {PropTypes} from '../react-compat'
+
+const getMockedContext = unsubscribe => ({
+  [CHANNEL]: {
+    getState: () => {},
+    setState: () => {},
+    subscribe: () => unsubscribe,
+  },
+})
+
+test('renders a non-glamorous component with theme', () => {
+  const CompWithTheme = withTheme(({theme: {padding}}) => (
+    <div style={{padding}} />
+  ))
+  expect(
+    render(
+      <ThemeProvider theme={{padding: '10px'}}>
+        <CompWithTheme />
+      </ThemeProvider>,
+    ),
+  ).toMatchSnapshot()
+})
+
+test('theme properties updates get propagated down the tree', () => {
+  class Parent extends Component {
+    state = {
+      padding: 10,
+    }
+
+    render() {
+      return (
+        <ThemeProvider theme={{padding: this.state.padding}}>
+          <Child />
+        </ThemeProvider>
+      )
+    }
+  }
+
+  const Child = withTheme(({theme: {padding}}) => <div style={{padding}} />)
+  const wrapper = mount(<Parent />)
+  expect(wrapper).toMatchSnapshot(`with theme prop of padding 10px`)
+  wrapper.setState({padding: 20})
+  expect(wrapper).toMatchSnapshot(`with theme prop of padding 20px`)
+})
+
+test('works properly with classes', () => {
+  /* eslint-disable react/prefer-stateless-function */
+  class Child extends Component {
+    render() {
+      const {theme: {padding}} = this.props
+      return <div style={{padding}} />
+    }
+  }
+
+  Child.propTypes = {
+    theme: PropTypes.object,
+  }
+
+  const ChildWithTheme = withTheme(Child)
+
+  class Parent extends Component {
+    state = {
+      padding: 10,
+    }
+
+    render() {
+      return (
+        <ThemeProvider theme={{padding: this.state.padding}}>
+          <ChildWithTheme />
+        </ThemeProvider>
+      )
+    }
+  }
+
+  const wrapper = mount(<Parent />)
+  expect(wrapper).toMatchSnapshot(`with theme prop of padding 10px`)
+})
+
+test('pass through when no theme provider found up tree', () => {
+  /* eslint-disable no-console */
+  const originalWarn = console.warn
+  console.warn = jest.fn()
+
+  const Child = withTheme(() => <div />)
+  const Parent = () => <Child />
+  const wrapper = mount(<Parent />)
+
+  expect(wrapper).toMatchSnapshot(`without theme provider`)
+  expect(console.warn).toHaveBeenCalledTimes(1)
+  expect(console.warn).toHaveBeenCalledWith(
+    // eslint-disable-next-line max-len
+    `glamorous warning: Expected component called "Stateless Function" which uses withTheme to be within a ThemeProvider but none was found.`,
+  )
+  console.warn = originalWarn
+})
+
+test('does not warn when NODE_ENV is set to `production`', () => {
+  /* eslint-disable no-console */
+  const originalWarn = console.warn
+  console.warn = jest.fn()
+
+  const originalEnv = process.env.NODE_ENV
+  process.env.NODE_ENV = 'production'
+
+  const Child = withTheme(() => <div />)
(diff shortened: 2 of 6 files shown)


kkemple:
Also, README needs updating still (and do you want an example?). I just wanted to wait until we were good from a code standpoint.

src/with-theme.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    componentWillUnmount() {
+      // cleanup subscription
+      this.unsubscribe && this.unsubscribe()
+    }
+
+    render() {
+      return (
+        <ComponentToTheme {...Object.assign({}, this.props, this.state)} />
```
kentcdodds: This could be simply:

```jsx
<ComponentToTheme {...this.props} {...this.state} />
```

src/__tests__/with-theme.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const Child = withTheme(({theme: {padding}}) => <div style={{padding}} />)
+  const wrapper = mount(<Parent />)
+  expect(wrapper).toMatchSnapshotWithGlamor(`with theme prop of padding 10px`)
+  wrapper.setState({padding: 20})
+  expect(wrapper).toMatchSnapshotWithGlamor(`with theme prop of padding 20px`)
+})
+
+test('acts as pass through when no theme provider found up tree', () => {
```
kentcdodds: Do we have any way of detecting whether there's a provider up the tree? If so, maybe we should log a warning?

src/__tests__/with-theme.js
```diff
@@ -0,0 +1,71 @@
+/* eslint func-style:0 */
+import React, {Component} from 'react'
+import {render, mount} from 'enzyme'
+import * as jestGlamorReact from 'jest-glamor-react'
```
kentcdodds: Because you're using the `style` prop to apply styles (which I think is actually really smart), you shouldn't need this. You can just use `toMatchSnapshot` :)

src/__tests__/with-theme.js
kkemple: i had a console.warn statement in at first but it broke linting (not that it can't be ignored). I don't think you can really tell if you have a provider, but you can definitely warn if there is no `__glamorous__` context prop (which is a good indicator of that situation).

src/__tests__/with-theme.js
kkemple: yeah i mean to update that :blush:

src/with-theme.js
kkemple: :man_facepalming:

src/__tests__/with-theme.js
kentcdodds: I think let's just leave it as it is for now... Maybe we'll add a warning in the future. Thanks!

kkemple:
Updated per your review comments, let me know if there is anything I missed. I added a `console.warn` statement if there is no `__glamorous__` context found. So be sure to give that a review as well!

src/__tests__/with-theme.js
kkemple: oops just saw this, will revert that lol

kkemple:
removed `console.warn` statement if no context found 😄

kkemple:
👍

kwelch:
Do you think there should be a test that ensures it works when wrapping a class?

Other than that, it looks good.

kentcdodds:
> Also, do you want any info about this new API in README?

Oh! Of course! Thanks for mentioning it. Yes, let's please have a section about this right under the docs about the `ThemeProvider` :+1:

src/with-theme.js
```diff
@@ -0,0 +1,38 @@
...
+    setTheme = theme => this.setState({theme})
+
+    componentWillMount() {
+      if (this.context[CHANNEL]) {
+        this.setState({theme: this.context[CHANNEL].getState()})
+      }
+    }
+
+    componentDidMount() {
+      if (this.context[CHANNEL]) {
```
vesparny: `withTheme()` is supposed to be used inside a `<ThemeProvider />`.
Shouldn't we throw an error here if `!this.context[CHANNEL]` ?

This would warn the user before he realize he does not have access to `theme` in children glamorous comps.

kentcdodds: We actually [discussed](https://github.com/paypal/glamorous/pull/45#discussion_r111036984) adding a warning (and @kkemple even implemented it before I said we should just leave it). Maybe we should add it back after all. I would want it to be wrapped in a `process.env.NODE_ENV === 'production'` call though. I'm uncertain whether our rollup config is currently capable of removing this for our UMD build. May need to investigate that a bit.

kkemple: I added this back in under a process.env if statement so it should be tree shakeable once the build part is figured back out. I had to ignore the console statement for eslint and the if statement for instanbul since there is no way to test that statement safely (that I am aware of).

kkemple:
I've added a section to the README and implemented a warning in production when there is no `ThemeProvider` further up the tree. :smile:

README.md
```diff
@@ -344,6 +344,72 @@ const Title = glamorous.h1({
 </ThemeProvider>
 ```
 
+`glamorous` also exports a `withTheme` higher order component (HOC) so you can access your theme in any component!
+> Try this out in your browser [here](https://codesandbox.io/s/qYmJjE4jy)!
```
kentcdodds: Very cool! When this gets merged and released, you'll need to update the glamorous version in your packages on that codesandbox :)

README.md
```diff
@@ -344,6 +344,72 @@ const Title = glamorous.h1({
...
+// extended component with theme prop
+const ThemedSubTitle = withTheme(SubTitle);
+
+<ThemeProvider theme={theme}>
+  <Title>Hello!</Title>
+  <ThemedSubTitle>from withTheme!</ThemedSubTitle>
+</ThemeProvider>
+```
+
+Or if you prefer decorator syntax:
```
kentcdodds: Didn't realize it worked like this out of the box. Neat!

README.md
```diff
@@ -344,6 +344,72 @@ const Title = glamorous.h1({
...
+    return <h3 style={{color}}>{children}</h3>;
+  }
+}
+
+<ThemeProvider theme={theme}>
+  <Title>Hello!</Title>
+  <SubTitle>from withTheme!</SubTitle>
+</ThemeProvider>
+```
+> `withTheme` expects a `ThemeProvider` further up the render tree and will warn in `production` if one is not found!
```
kentcdodds: Should only warn in `development` :+1:

src/with-theme.js
```diff
@@ -0,0 +1,46 @@
...
+
+export default function withTheme(ComponentToTheme) {
+  class ThemedComponent extends Component {
+    state = {theme: {}}
+    setTheme = theme => this.setState({theme})
+
+    componentWillMount() {
+      if (!this.context[CHANNEL]) {
+        /* istanbul ignore if */
+        if (process.env.NODE_ENV === 'production') {
```
kentcdodds: This should be: `if (process.env.NODE_ENV !== 'production') {`

We don't want to warn in production :)

src/with-theme.js
```diff
@@ -0,0 +1,46 @@
...
+          /* eslint-disable no-console */
```
kentcdodds: Could we use this instead:

```
// eslint-disable-next-line no-console
```

src/__tests__/with-theme.js
```diff
@@ -0,0 +1,101 @@
...
+
+  const wrapper = mount(<Parent />)
+  expect(wrapper).toMatchSnapshot(`with theme prop of padding 10px`)
+})
+
+test('pass through when no theme provider found up tree', () => {
+  const Child = withTheme(() => <div />)
+  const Parent = () => <Child />
+  const wrapper = mount(<Parent />)
+  expect(wrapper).toMatchSnapshot(`without theme provider`)
```
kentcdodds: Could we add an assertion here that `console.warn` was called? We'd need to mock `console.warn`:

```javascript
const originalWarn = console.warn
console.warn = jest.fn()
// your test code you have already
expect(console.warn).toHaveBeenCalledTimes(1)
expect(console.warn). toHaveBeenCalledWith('Expected ThemeProvider but none found!')
```

And actually, if we could make that warning a little more helpful that'd be great. Like providing `ComponentToTheme .displayName || ComponentToTheme .name` and saying it's coming from glamorous. So like:

```javascript
// eslint-ignore-next-line max-len
console.warn(`glamorous warning: Expected component called ${ComponentToTheme .displayName || ComponentToTheme .name} which uses withTheme to be within a ThemeProvider but none was found.`)
```

I don't mind being verbose about this :smile:

README.md
kkemple: oh oops, i thought you said in production, will update :smile:

src/with-theme.js
kkemple: yeah we can for sure

src/__tests__/with-theme.js
kkemple: I figured there would be some adjustment there, just filled in anything until you figured out warning message.

src/with-theme.js
kkemple: did not know about `eslint-disable-next-line` much better!

README.md
kkemple: oh right, good call!

README.md
kkemple: Yeah I haven't looked back since I discovered that! :smile:
I have great one that is for checking internet connection `@connection` adds an offline prop that is a boolean. It's great.

kkemple:
@kentcdodds made requested changes!

src/with-theme.js
```diff
@@ -0,0 +1,57 @@
+import React, {Component} from 'react'
+
+import {CHANNEL} from './theme-provider'
+import {PropTypes} from './react-compat'
+
+function generateWarningMessage(componentName) {
+  // eslint-disable-next-line max-len
+  return `glamorous warning: Expected component called ${componentName} which uses withTheme to be within a ThemeProvider but none was found.`
```
kentcdodds: Why don't we wrap `componentName` this in quotes?

src/with-theme.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export default function withTheme(ComponentToTheme) {
+  class ThemedComponent extends Component {
+    state = {theme: {}}
+    setTheme = theme => this.setState({theme})
+
+    componentWillMount() {
+      if (!this.context[CHANNEL]) {
+        // istanbul ignore else
```
kentcdodds: Actually, I'd really like to make sure that we don't log in production. So let's add a test that basically does:

```javascript
const originalEnv = process.env.NODE_ENV
process.env.NODE_ENV = 'production'

// tests

process.env.NODE_ENV = originalEnv
```

Thanks!

kkemple: added!

src/with-theme.js
kkemple: done!

src/__tests__/with-theme.js
```diff
@@ -0,0 +1,130 @@
...
+  console.warn = jest.fn()
+
+  const originalEnv = process.env.NODE_ENV
+  process.env.NODE_ENV = 'production'
+
+  const Child = withTheme(() => <div />)
+  const Parent = () => <Child />
+  mount(<Parent />)
+
+  expect(console.warn).toHaveNotBeenCalled
```
kentcdodds: This won't work I'm afraid. Should be:

```js
expect(console.warn).not.toHaveBeenCalled()
```

kkemple: oh right! lol

kkemple:
okay, updated!

kentcdodds:
@kkemple, thanks so much for your help! I've added you as a collaborator on the project. Please make sure that you review the CONTRIBUTING.md file (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

kkemple:
Thanks! I look forward to contributing and helping others do the same! :smile:

PR #46, opened 2017-04-12
Add withTheme higher order component

tkh44:
**What**: Add withTheme higher order component

**Why**: Enables developers to use themes while using glamorous.Div.

**How**: created withTheme file and tests.

This is really just a proof of concept to get a feel for interest, albeit a working and documented one 😄 . I need to get 100% coverage still. Since most of the code is copy/pasted from `GlamorousComponent` I can just use those tests as guides.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -140,6 +140,13 @@
       "contributions": [
         "review"
       ]
+    },
+    {
+      "login": "tkh44",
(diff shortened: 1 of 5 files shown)


tkh44:
#45 is a much more comprehensive PR. 👍

kentcdodds:
Thanks for this anyway!

Issue #23, opened 2017-04-06
use html-element-attributes


kentcdodds:
Right now we've got our own list of html-element-attributes. But we should probably rely on another package that will keep itself updated. So it was suggested in #8 that we use https://github.com/wooorm/html-element-attributes from @wooorm

It doesn't have all the svg attributes (maybe that's something @wooorm would be willing to support?). And it also doesn't have the react-specific props so we'll probably want to keep those too. But for some of this we could simplify [this file](https://github.com/paypal/glamorous/blob/f75fd4556edf826c927e6ffba20a7f950229b133/src/should-forward-property.js) a bit by using that. Pretty much just need a refactor of that file to replace `htmlProps` with that file's stuff. One thing that `shouldForwardProperty` accepts is a `tagName` which is great because it means we can leverage the knowledge of tag-specific props that html-element-attributes has to be more specific about which props to forward.

I think we should probably add some unit tests for that file as well so it's more clear what that module is capable of/responsible for.

If anyone wants to take this on, let us know!

jackyho112:
Hello @kentcdodds, I have been following you since I watched your webpack course. I also read your blog post and watched your videos on open source project contribution. 

I do want to contribute to open source projects much more often than how much I am doing now. I believe this issue is a good opportunity and I think I can handle it. With that said, may I give this issue a go?

kentcdodds:
Absolutely! Let me know how I can help!

jackyho112:
Thanks! 

For checking html props, do you want me to take into account the tag name? That means `isHtmlProp` will only check if the property name exists in the global attributes and the attributes that belong to that tag in `html-element-attributes`. If so, I will have to add more attributes to `svgProps` because `html-element-attributes` does not include SVG as a tag and some SVG attributes are missing in `svgProps`.

Otherwise, I will simply set `htmlProps` to all the html attributes in `html-element-attributes` and everything else will stay the same.

kentcdodds:
Yes, that's actually one of the benefits of going with html-element-attributes is that the attributes are scoped to the tag name and we have that (most of the time) so we can do a better job of determining which props to forward to the underlying element. So, to address your comments/questions:

> That means isHtmlProp will only check if the property name exists in the global attributes and the attributes that belong to that tag in html-element-attributes.

Perfect :+1:

>  I will have to add more attributes to svgProps because html-element-attributes does not include SVG as a tag and some SVG attributes are missing in svgProps.

Also perfect :+1:

Thanks!

[...]

PR #47, opened 2017-04-12
Implement react-html-attributes 

jackyho112:
**What**:
Refactored `should-forward-property.js` to use `react-html-attributes` which is a package that is essentially a store with react-supported html attributes keyed by their respective html tags. 

So instead of checking among all the html attributes to decide whether to forward a property or not, we will be using the react-supported versions of these attributes. 

`react-html-attributes` will also include SVG attributes supported by react, so there will be no need to keep a copy of SVG properties in this library like it does currently.

Also added tests for `should-forward-property.js`.

**Why**:
https://github.com/paypal/glamorous/issues/23

To summarize the discussion in that link regarding the issue, these changes are meant to 
1. delegate the construction and delivery of the html attributes including SVG attributes supported by React to an external library which ideally will also map the attributes to their respective html tags
2. make sure `should-forward-property.js` maintains its core functionality through future changes and is made more clear with what it's trying to achieve with tests

**How**:
1. remove the html attribute and SVG attribute lists from `should-forward-property.js`.
2. get those attributes keyed by their tags from `react-html-attributes`
3. write unit tests for `should-forward-property.js` to test for all its core use cases with some examples regarding its main predicate function

https://github.com/jackyho112/react-html-attributes - this package is written specifically for the purpose of this branch. Feel free to also review this to check if it fulfills what it needs to do.


paulmolluzzo:
> What is the impact in terms of size to the final glamorous build?

Good question @vesparny.

Can our code coverage report include this? Maybe there's some other CI package/lib that can do this calculation on new PRs?

.all-contributorsrc
```diff
@@ -140,6 +140,15 @@
...
         "review"
       ]
+    },
+    {
+      "login": "jackyho112",
+      "name": "jackyho112",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/25375401?v=3",
+      "profile": "https://github.com/jackyho112",
+      "contributions": [
+        "code"
```
kentcdodds: You're selling yourself short :wink: you also contributed tests! If you would like, please add `"test"` to this array then update the contributor table as documented in the CONTRIBUTING.md 😄

package.json
```diff
@@ -27,7 +27,8 @@
   "dependencies": {
     "brcast": "^1.1.6",
     "html-tag-names": "^1.1.1",
-    "svg-tag-names": "^1.1.0"
+    "svg-tag-names": "^1.1.0",
+    "react-html-attributes": "^1.0.2"
```
kentcdodds: Any chance you could update the `yarn.lock` file as well? You'll need [`yarn`](https://yarnpkg.com) installed, and then you just run `yarn` and it should update itself.

src/should-forward-property.js
```diff
@@ -600,17 +26,18 @@ const isCustomAttribute = RegExp.prototype.test.bind(
 )
 
 const hasItem = (list, name) => list.indexOf(name) !== -1
-const isHtmlProp = name => hasItem(htmlProps, name)
+const isHtmlProp = (name, tagName) => {
+  const elementAttributes = reactHTMLAttributes[tagName]
+
+  return hasItem([...globalReactHtmlProps, ...elementAttributes], name)
```
kentcdodds: Because this code is run a lot, I want to be very considerate of performance here. Could we write this function like this instead:

```javascript
function isHtmlProp(name, tagName) {
  const elementAttributes = reactHTMLAttributes[tagName]
  // elementAttributes could be undefined, so we do defined check here
  return hasItem(globalReactHtmlProps, name) || elementAttributes === undefined ? false : hasItem(elementAttributes, name)
}
```

And actually, while we're here, we could probably speed this up a little bit by removing the `hasItem` helper function and just use `x.indexOf(y) !== -1` everywhere. Am I over-optimizing? Hard when we don't really have benchmarking set up. I'm just freaked out by #43 :wink: and this is definitely "hot code" so I want to make sure it's fast :)

src/should-forward-property.js
```diff
@@ -600,17 +26,18 @@ const isCustomAttribute = RegExp.prototype.test.bind(
...
-const isSvgProp = (tagName, name) =>
-  tagName === 'svg' && hasItem(svgProps, name)
 const isReactProp = name => hasItem(reactProps, name)
 
 // eslint-disable-next-line complexity
 const shouldForwardProperty = (tagName, name) =>
   typeof tagName !== 'string' ||
-  ((isHtmlProp(name) ||
-    isSvgProp(tagName, name) ||
+  ((isHtmlProp(name, tagName) ||
```
kentcdodds: Glad to combine svg and html props here :+1:

jackyho112:
Thanks for all the comments. I will make a fix to this branch to account for them!

src/should-forward-property.js
jackyho112: @kentcdodds 

I will implement those. And I have some other ideas that may make this faster. 

Will putting `isReactProp(name)` before `isCustomAttribute(name)` make it faster? Because I think the former could be a more common use case. On a related note, how about putting `!isCssProp(name)` before `tagName === 'svg'`? Since I think the former may turn out to be true more often than the latter. 

Another idea is caching the result of `shouldForwardProperty` in its file by using closure so that the combination of `tagName` and `name` that has already been checked before doesn't need to go through the same process again to get the result.

kentcdodds: Actually the `tagName === 'svg'` lookup should be extremely fast. So I'd leave that where it is. Also you're probably right, putting the `isReactProp` before the `isCustomAttribute` would be faster. So we can swap those if you like. Hard to tell really without some hard benchmarks though so 🤷‍♂️

kentcdodds: > Another idea is caching the result of shouldForwardProperty in its file scope by using closure so that the combination of tagName and name that has already been checked before doesn't need to go through the same process again to get the result.

Oh yeah, and memoization is what you're talking about and that would probably be useful! I don't know if there's a super small memoization library, but might be worth looking into before implementing it ourselves :)

kentcdodds: Note that some of these optimizations may not be necessary for this PR, so feel free to put some of those off to some other time :)

jackyho112: Found one.

https://www.npmjs.com/package/fast-memoize

It claims to be 'fastest possible memoization library in JavaScript'. And it's 616B gzipped. 

If you want this to be implemented, maybe I can just open another branch and pr to add this?

kentcdodds: And I think that gzip size isn't even minified. That's awesome. Looks perfect. I say it's worth the bytes :+1:

Feel free to implement it as part of this PR or a new one 👍

README.md
```diff
@@ -3,7 +3,7 @@
 # glamorous
 
 React component styling solved with an elegant ([inspired](#inspiration)) API,
-small footprint (~6kb gzipped), and great performance (via [`glamor`][glamor]).
+small footprint (~7kb gzipped), and great performance (via [`glamor`][glamor]).
```
kentcdodds: Maybe we should just remove the actual number. It seems to be fluctuating a bit anyway.

jackyho112: I will do that and implement https://www.npmjs.com/package/fast-memoize.

kentcdodds:
Hey @jackyho112 , thanks so much for your help! I've added you as a collaborator on the project. Please make sure that you review the CONTRIBUTING.md file (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

jackyho112:
@kentcdodds You are very welcome! Glad to have helped! I will definitely keep an eye on this repo and look forward to making more contribution.

Issue #50, opened 2017-04-13
Typings are not published


sammkj:

kentcdodds:
Oh yeah, we'll probably need to update this: https://github.com/paypal/glamorous/blob/master/package.json#L16

PR #51, opened 2017-04-13
Update typescript typings to be stricter

luke-john:
**What**:
Made the typescript definitions stricter so when adding css props to a html element the compiler will complain on misspellings and incorrect usage of props.

**Why**:
Because it gives confidence that styles are used correctly and gives you hints about what properties are valid.

image

**How**:
Using the Definitions from react around what are valid style properties for HTML and SVG elements.


kwelch:
This might be a good PR to take care of #50, if you understand how.

typings/glamorous.d.ts
```diff
@@ -3,137 +3,33 @@
 // Definitions by: Kok Sam <https://github.com/sammkj>
 
 import * as React from "react";
-import { StatelessComponent, ComponentClass, PureComponent } from "react";
+import { ComponentClass, CSSProperties, StatelessComponent, SVGAttributes } from 'react'
```
sammkj: Can you change single quote to double quote?

kentcdodds: Actually, we should follow the standard of the project. Could you use single quotes only please?

sammkj: Ahh sorry, I didn't notice!

luke-john: updated initial react import to use singles as well.

luke-john:
Because I've split the typings into multiple files this pr would be broken by https://github.com/paypal/glamorous/pull/51.

A couple of possible options are
1. I can update this pr to have the definitions in a single file.
2. #51 could be updated to copy the `.d.ts` files to a typings folder under dist.
3. Hold off on landing this pr until 51 lands and I update this pr to include the solution in option 2.

kentcdodds:
Thanks @luke-john! I think we should probably just publish the `typings` directory instead. Is there any way we could move everything except `.d.ts` files out of that directory? We could add a `test` directory for the TypeScript test file. Does that sound ok? If you could update the `files` property in the `package.json` to include `typings`, and make those other changes, I think we can close #54 and just use this one :+1:

luke-john:
I've updated the structure as following in an additional commit

```
/test
  - glamorous.test.ts
/typings
  - ...definition files
- tsconfig.json
- tslint.json
```

kentcdodds:
I'd like for @sammkj to review the changes, then we can merge it :)

sammkj:
@kentcdodds LGTM! Let's merge this

PR #52, opened 2017-04-13
fix(docs): fix `withTheme` example, add kkemple to contributors

kkemple:
**What**:
Fixed an error in the docs for `withTheme` example. I fixed in code sandbox example and wanted to reflect the correct example here as well. Also added myself to contribs since I forgot to add to `withTheme` PR.

**Why**:
If a user copied the example for `withTheme` it would not work properly.

**How**:
Updated README

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -150,6 +150,17 @@
         "code",
         "test"
       ]
+    },
+    {
+      "login": "kkemple",
+      "name": "Kurtis Kemple",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/3629876?v=3",
+      "profile": "https://twitter.com/kurtiskemple",
+      "contributions": [
+        "code",
+        "test",
+        "doc"
+      ]
     }
   ]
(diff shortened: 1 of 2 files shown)


.all-contributorsrc
```diff
@@ -150,6 +150,16 @@
...
       ]
+    },
+    {
+      "login": "kkemple",
+      "name": "Kurtis Kemple",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/3629876?v=3",
+      "profile": "https://twitter.com/kurtiskemple",
+      "contributions": [
+        "code",
+        "doc"
```
kentcdodds: Want to add test here too? See the contributing guidelines for how to regenerate after you update this 😉 if your want to.

kkemple: oh, i missed test, yeah, i'll update!

kkemple:
thanks! 🎉

kentcdodds:
Thank YOU!

PR #54, opened 2017-04-13
fix(typings): copy glamorous.d.ts to the dist directory

### LINKED ISSUE #50: Typings are not published

kentcdodds:
**What**: This adds a script to copy the d.ts file to the `dist`
directory and updates the `package.json` to point to that file.

**Why**: So it's published with the rest of the stuff.

**How**: Using `npsUtils.copy` and update the default build script

Closes #50

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "dist/glamorous.cjs.js",
   "jsnext:main": "dist/glamorous.es.js",
   "module": "dist/glamorous.es.js",
-  "typings": "typings/glamorous.d.ts",
+  "typings": "dist/glamorous.d.ts",
   "scripts": {
     "start": "nps",
     "test": "nps test",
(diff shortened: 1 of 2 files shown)


kentcdodds:
Anyone can merge this as soon as we get a green build. No need to change the commit message.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "dist/glamorous.cjs.js",
   "jsnext:main": "dist/glamorous.es.js",
   "module": "dist/glamorous.es.js",
-  "typings": "typings/glamorous.d.ts",
+  "typings": "dist/glamorous.d.ts",
```
kentcdodds: I preferred to do it this way rather than just publish the `typings` directory because I like the clean separation between the source code and the published code.

kentcdodds:
Actually, based on #51, I'm thinking that we should probably just publish the `typings` directory instead.... Let me respond to that PR.

PR #56, opened 2017-04-14
fix(typings): Update Typescript definitions

sammkj:
**What**:

Some of the TS definitions don't match with current implementation. For example, `glamorous` takes in
any number of style object and function. But the type definition only allows two arguments. Also,
definition for `glamorous` returns incorrect type. This PR also adds definition for `ThemeProvider`

This PR is built upon #51.


sammkj:
@kentcdodds is there anyone else maintaining the TS definitions?

kentcdodds:
Nobody that I'm aware of. I want to get #51 to merge before doing anything else with all this stuff. Once we actually get these things published, we can start making improvements.

sammkj:
Closing this as I will create a new PR

PR #57, opened 2017-04-14
fix(readme): Improve composition docs and add missing examples

### LINKED ISSUE #55: Improve documentation for component composition and fix `css` prop docs
I've been using `glamorous` in a few test applications and got a bit stuck when trying to work out the best way of doing component composition. I couldn't find anything related to this in the documentation so thought about opening this "issue". 

- `glamorous` version: 3.5.0
- `glamor` version: 2.20.24
- `react` version: 15.5.4

**Reproduction:**

https://codesandbox.io/s/OyM7DoxLg

**Problem description:**

I couldn't find an easy way of using the styles from a `glamorous` component without having to create a functional component. Something like `styled-components` does with the [styled()](https://github.com/styled-components/styled-components/blob/master/README.md#overriding-component-styles) helper.

Is this actually possible to do at the moment?

**Suggested solution:**

In the sandbox I suggested a couple of options, but I'm not sure which would be the best.

Doing something like 
```js
const LargeText = glamorous.div([
   Text,
   { fontSize: 24 }
]);
```
would give the flexibility of using a different HTML element if needed, which might be useful.

But going on the `styled-components` route of using something like
```js
const LargeText = withStyles([Text, { fontSize: 24 }]);
```
keeps things simple. I used `withStyles` in line with `withTheme` but alternatively it could maybe be simply `glamorous([Text, { fontSize: 24 }])` since the HTML element would be inferred from the `Text` component. This could have some issues when composing more than one component so not sure about that.

**Side note**

I think I might have found a bug with `cssOverrides` in the sandbox I posted above. If that's the case I can open a separate issue if you want, but maybe I'm just expecting a different behaviour.

Thanks!

braposo:
**What**:
- Updates `README.md` with examples about overriding styles with composition
- Updates `cssOverrides` documentation to use the right `css` prop instead
- Adds other examples for component factories

**Why**:
It isn't clear in the documentation how you can use `glamorous()` to override existing components' styles through composition. While searching for that, I (read @kentcdodds) found that `cssOverrides` was still mentioned in the docs when the public prop is actually `css`.

**How**:
I started by adding a separate section for overriding the component styles through composition, but then realised that it made sense to also include `css` and `className` because they are used for the same thing so people will look for that in there.

Since I didn't want to repeat what was already in the "GlamorousComponent" section, I thought about adding just the `glamorous()` example below the other two but also realised it didn't make much sense as it's not really about a "GlamorousComponent" prop like `css` and `className`.

So my decision was to move the `css` and `className` sections from the "GlamorousComponent" section to the (new) "Overriding component styles" section, where I also put the `glamorous()` example. Because it's still important to mention those props in the "GlamorousComponent" section, I added a new paragraph about supported props and directed people to both the "Overriding component styles" and "Theming" sections in case they want to know more.

Hope this explanation (and the changes) make sense to you too, but (really) feel free to suggest anything that might be better suited.

I also added a few other examples to sections that weren't totally clear to me when I first read the docs so took the opportunity to add those as well. It will help people (like me) that need an example to fully understand what the text means.

Closes #55


README.md
```diff
@@ -174,67 +182,52 @@ wrapping a component you intend to style, you'll need to make sure you accept
 the `className` as a prop and apply it to where you want the styles applied in
 your custom component (normally the root element).
 
+```javascript
+const UnstyledComp = ({ className, children }) => <div className={`${className} other-class`>{children}</div>;
+const MyStyledComp = glamorous(UnstyledComp)({ margin: 1 });
+
+render(<MyStyledComp>content</MyStyledComp>);
+
+// rendered output: <div class="<glamor generated id> other-class">content</div>
```
kentcdodds: All the same comments for this example too :+1:

README.md
```diff
@@ -174,67 +182,52 @@ wrapping a component you intend to style, you'll need to make sure you accept
 the `className` as a prop and apply it to where you want the styles applied in
 your custom component (normally the root element).
 
+```javascript
+const UnstyledComp = ({ className, children }) => <div className={`${className} other-class`>{children}</div>;
+const MyStyledComp = glamorous(UnstyledComp)({ margin: 1 });
```
kentcdodds: Also, let's follow the convention of the codebase and not use semicolons anywhere.

README.md
```diff
@@ -174,67 +182,52 @@ wrapping a component you intend to style, you'll need to make sure you accept
...
-const myCustomGlamorStyles = glamor.css({fontSize: 2, padding: 2})
-<MyStyledDiv
-  className={`${myCustomGlamorStyles} custom-class`}
-  cssOverrides={{padding: 3}}
-/>
-// styles applied:
-// {margin: 1, fontSize: 2, padding: 3}
-// as well as any styles custom-class applies
-```
+By default `GlamorousComponent` supports three props: `className`, `css` and `theme`
```
kentcdodds: Perfect :+1:

README.md
```diff
@@ -291,6 +284,71 @@ One other tip... This totally works:
 </glamorous.Div>
 ```
 
+#### Overriding component styles
```
kentcdodds: Above it's mentioned that the three ways props are `css`, `className`, and `theme`. Despite the fact that `theme` is documented elsewhere, I think it would be valuable to add a little documentation here as well (even if it's just a mention).

README.md
```diff
@@ -291,6 +284,71 @@ One other tip... This totally works:
...
+
+The most common scenario for using props is to override the style of an
+existing component (generated by `glamorous` or not). In this section we'll
+explain how you can use `className`, `css` and composition to override the
+styles of a component. Let's see how that can be done in the examples below.
+
+> Try this out in your browser [here](https://codesandbox.io/s/Kro5369wG)!
```
kentcdodds: On your example, there's:

```jsx
// Adds baseStyles to `UnstyledTextNoClassName` through `glamorous`
// (this won't work)
const ComposedUnstyledTextNoClassName = glamorous(UnstyledTextNoClassName)(
  baseStyles,
);
```

Could you change this to:

```jsx
// Adds baseStyles to `UnstyledTextNoClassName` through `glamorous`
// (this won't work because the underlying component doesn't apply the
// className that it's given from glamorous).
const ComposedUnstyledTextNoClassName = glamorous(UnstyledTextNoClassName)(
  baseStyles,
);
```

kentcdodds: Also, because this is what we're showing that people _should_ be doing, could you change this:

```jsx
// Simulates existing component
const UnstyledText = ({ className, children }) => (
  <div className={className}>{children}</div>
);
```

To this:

```jsx
// Simulates existing component
// Note that the most important parts of this component
// is that it will forward the children and className props
// and it will also forward any other valid div attributes
// like `id` and `data-`/`aria-` attributes
const UnstyledText = (props) => (
  <div {...props} />
);
```

Maybe that's more confusing than anything... I'm going to leave that up to you. But I want to avoid people asking us the question: "How do I make it so I can pass an `id` to my underlying component" which is why I brought it up. Maybe an additional comment would be sufficient...

Could also just do:

```jsx
// Simulates existing component
const UnstyledText = ({ className, children, ...rest }) => (
  <div className={className} {...rest}>{children}</div>
);
```

That's more work technically, but it at least indicates the importance of `className` and `children`. Your call :)

README.md
```diff
@@ -291,6 +284,71 @@ One other tip... This totally works:
...
+
+We'll use this as our `GlamorousComponent`:
+```javascript
```
kentcdodds: This one's fine to be `javascript` because there's no `jsx`. Could you put a blank line above this one though? I like to give all my code fences some space.

README.md
```diff
@@ -291,6 +284,71 @@ One other tip... This totally works:
...
+
+```jsx
+const MyComposedStyledDiv = glamorous(MyStyledDiv)({fontSize: 4, padding: 4})
+<MyComposedStyledDiv />
+// styles applied:
+// {margin: 1, fontSize: 4, padding: 4}
+```
+
+In fact, the built-in DOM component factories provided are just an abstraction
+of this function so `glamorous.div` could be written as `glamorous("div")` instead.
```
kentcdodds: Perfect. Could you change `glamorous("div")` to `glamorous('div')`?

README.md
kentcdodds: Oh, one other thing on your codesandbox, make sure to change the name and the description so it's not "glamorous help" but it's "glamorous example" or something more descriptive :)

README.md
```diff
@@ -163,6 +163,14 @@ For every DOM element, there is an associated `glamorous` component factory
 attached to the `glamorous` function. As above, you can access these factories
 like so: `glamorous.div`, `glamorous.a`, `glamorous.article`, etc.
 
+```javascript
```
braposo: 👍 https://github.com/braposo/glamorous/commit/a3f14141c9a2090095934c324f9e1e096d9edb91

README.md
```diff
@@ -163,6 +163,14 @@ For every DOM element, there is an associated `glamorous` component factory
...
+const MyStyledSection = glamorous.section({ margin: 1 })
+
+render(<MyStyledSection>content</MyStyledSection>);
```
braposo: 👍 https://github.com/braposo/glamorous/commit/bafd9cba82f0936b20d740cf4d25d7149afb05c4

README.md
```diff
@@ -163,6 +163,14 @@ For every DOM element, there is an associated `glamorous` component factory
...
+
+// rendered output: <section class="<glamor-generated-id>">content</section>
```
braposo: 👍 https://github.com/braposo/glamorous/commit/5e608f4810bb708d0e30f8474c2b874ffaf5fbfc

README.md
braposo: I'm just too used to that 🙈 https://github.com/braposo/glamorous/commit/cbbde3016e342899d014279688aaeb7d32a8d94b

Also found a couple of other places with semis so removed them as well.

README.md
braposo: done in previous commits

README.md
braposo: Yup makes sense https://github.com/braposo/glamorous/commit/4af3a70bcafbd518978c5a2a8be74770ce083196

README.md
braposo: I prefer the last option in your second comment. I don't like to see `{...props}` in a component because of the likelihood of triggering unknown props warnings, not that `{...rest}` wouldn't cause it but at least by destructuring the `props` it might give the hint that it could be sanitized if needed. Added a few more lines in the comment so it (hopefully) makes it clear that `className` is the important bit and that any other props are also supported through `{...rest}`.

Made the other suggested changes as well 👍

README.md
braposo: done in a previous commit 👍

README.md
braposo: also too used to double quotes 🙈

braposo: https://github.com/braposo/glamorous/commit/8c8c12e881488325a50b69240ef4b08706f4b5fa

braposo:
@kentcdodds I made your suggested changes and updated the sandbox as well. Please have a look and let me know if there's anything else I need to change.

Thanks for your words, I'm happy to help whenever it's possible and this looked like a good opportunity to improve the documentation.

Issue #59, opened 2017-04-14
Add support for `innerRef`


ElliotChong:
Instead of having to convolutedly wrap components to gain access to the inner element / component of a Glamorous-wrapped node it would be nice to have a convenience prop such as `innerRef` that would proxy down to the underlying element's `ref` property.

i.e.:
```
import { Div } from 'glamorous'

<Div innerRef={(node) => { console.log(node) }} />
```

kentcdodds:
I think I'm on board with this. Certainly enough to review a [pull request](http://makeapullrequest.com) :)

ElliotChong:
Thanks @rrdelaney and @kentcdodds ! :)

I was planning on doing a PR this weekend but was pleasantly surprised to see this already landed, you guys rock!

PR #60, opened 2017-04-15
feat(innerRef): Adds `innerRef` prop to access react element reference

rrdelaney:
This adds the optional `innerRef` prop to all glamorous component. `innerRef` will call a provided
function with the internal react element reference.

#59 

**What**: Adds an `innerRef` prop to easily access the react element reference

**Why**: Feature request in #59

**How**: Forward a received `innerRef` prop to React's `ref`


kentcdodds:
Looks great! Thank you!

PR #64, opened 2017-04-18
feat(perf): improve performance of GlamorousComponent#render

kentcdodds:
**What**: This takes some of the ideas from @developit to make the glamorous component render method faster

**Why**: Because that's some hot code!

**How**: copy/paste and blindly looking for other things I figured could be faster...

Ref #43

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,22 +3,30 @@ import commonjs from 'rollup-plugin-commonjs'
 import nodeResolve from 'rollup-plugin-node-resolve'
 import json from 'rollup-plugin-json'
 import uglify from 'rollup-plugin-uglify'
+import replace from 'rollup-plugin-replace'
 
 const minify = process.env.MINIFY
 const format = process.env.FORMAT
 const esm = format === 'es'
+const umd = format === 'umd'
+const cjs = format === 'cjs'
 
 let targets
 
-if (minify) {
-  targets = [{dest: 'dist/glamorous.umd.min.js', format: 'umd'}]
-} else if (esm) {
+if (esm) {
   targets = [{dest: 'dist/glamorous.es.js', format: 'es'}]
+} else if (umd) {
+  if (minify) {
+    targets = [{dest: 'dist/glamorous.umd.min.js', format: 'umd'}]
+  } else {
+    targets = [{dest: 'dist/glamorous.umd.js', format: 'umd'}]
+  }
+} else if (cjs) {
+  targets = [{dest: 'dist/glamorous.cjs.js', format: 'cjs'}]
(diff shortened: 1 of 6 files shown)


rollup.config.js
```diff
@@ -3,22 +3,30 @@ import commonjs from 'rollup-plugin-commonjs'
 import nodeResolve from 'rollup-plugin-node-resolve'
```
kentcdodds: Pretty much all the changes in this file are to make sure that `process.env.NODE_ENV === 'production'` gets swapped only in the UMD builds.

src/__tests__/index.js
```diff
@@ -213,6 +213,41 @@ test('renders a component with theme properties', () => {
   ).toMatchSnapshotWithGlamor()
 })
 
+test('in development mode the theme is frozen and cannot be changed', () => {
```
kentcdodds: This is technically a breaking change, but manipulation was never truly supported (or useful) so I'm going to let this one slide.

src/index.js
```diff
@@ -72,30 +72,37 @@ function glamorous(comp, {rootEl, displayName, forwardProps = []} = {}) {
...
 
       render() {
-        const {className, innerRef, ...rest} = this.props
-        const {toForward, cssOverrides} = splitProps(rest, GlamorousComponent)
+        // in this function, we're willing to sacrafice a little on
+        // readability to get better performance because it actually
+        // matters.
+        const props = this.props
+        const {toForward, cssOverrides} = splitProps(props, GlamorousComponent)
+        // freeze the theme object in dev mode
```
kentcdodds: Just a handy thing for dev mode so we don't have to make a copy of the object in production mode.

src/index.js
```diff
@@ -72,30 +72,37 @@ function glamorous(comp, {rootEl, displayName, forwardProps = []} = {}) {
...
           glamorStyles: parentGlamorStyles,
           glamorlessClassName,
-        } = extractGlamorStyles(className)
+        } = extractGlamorStyles(props.className)
         const glamorClassName = css(
           ...mappedArgs,
           ...parentGlamorStyles,
           cssOverrides,
         ).toString()
-        const fullClassName = joinClasses(glamorlessClassName, glamorClassName)
```
kentcdodds: I figured that calling out to a function that does all the stuff `joinClasses` did was unnecessary weight when the same could be done with simple string concat and `trim`.

paulmolluzzo:
What is the best way to get a performance evaluation that compares this PR to the way it was before?

developit:
@paulmolluzzo ES Bench CI, but it doesn't exist yet (sorry)

Issue #34, opened 2017-04-09
Add documentation for adding examples


kentcdodds:
I think that @pksjce may be the best one to do this considering she has first-hand experience! What do you say @pksjce?

pksjce:
Sure! Will take this up 👍

pksjce:
Hey, heads-up, Will be adding this as a piece in `CONTRIBUTING.MD`. Cool?

PR #65, opened 2017-04-18
docs(examples): Add documentation to create examples

pksjce:
As we recently agreed on the format to deal with examples for glamorous, this will document it.

BREAKING CHANGE: None

#34


kentcdodds:
Thank you so much! This is great!

PR #68, opened 2017-04-19
docs(examples): add documentation for ampersand and css combinators

mtcrutch:
**What**:
Add documentation and an example of `&` usage with CSS combinators.

**Why**:
Usage wasn't super clear to the consumer before.

**How**:
Via [Twitter](https://twitter.com/_jayphelps/status/854484171938381824). Tweet tweet.


mtcrutch:
Fixed!

Hit `enter` rather than `space` during the prompt 😁

examples/README.md
```diff
@@ -51,6 +51,31 @@ const MyGrid = glamorous.div({
 })
 ```
 
+## Ampersands & CSS Combinators
+
+You can use an ampersand (`&`) as a reference to the parent selector. This approach can be useful for CSS combinators: adjacent sibling selectors (`+`), general sibling selectors (`~`) and child selectors (`>`).
```
kentcdodds: Instead of "parent selector" I think we should describe it as "generated selector" or "current selector"

mtcrutch: Good call; that's a lot cleaner. Updating with "generated selector."

kentcdodds:
This is so great, thank you very much!

Issue #42, opened 2017-04-11
Make glamorous smaller


kentcdodds:
Because as @philpl [put it](https://twitter.com/_philpl/status/851639879628988422):

> SC v2 will be 11kb gzip, 32kb raw - although lower with preprocessing. @kentcdodds let the bytes war begin 😂

So here are a few ideas on the top of my head:

1. A lot of the size comes from the list of html attributes and element tag names. I'm not sure how SC is dealing with this but we may be able to learn from it. We could potentially make a build that skips the element tag names altogether and people could just do: `glamorous('div')({/* styles */})` on their own. It's mostly a convenience anyway... But I'm pretty sure I'd rather not do this... Other ideas? Maybe we could generate them at runtime somehow?
2. We added quite a bit with the `ThemeProvider` and not everyone needs to support theming. So we _could_ make a build that excludes the `ThemeProvider` related code 🤔 Things could get pretty messy though!
3. Anything else? Someone wanna peruse the code and see where we could shave some bytes?

threepointone:
I notice you're missing an es6 'module' build? so consumers can DCE efficiently.

threepointone:
(I'm doing the same for glamor)

kentcdodds:
Thanks for looking into it. We actually do have one: https://unpkg.com/glamorous@3.1.0/dist/glamorous.es.js :tada:

kentcdodds:
Also, I freaking LOVE rollup. Everytime I open up the rollup built file and see that it's totally flat I cheer: "Thank you @Rich-Harris" 👏

Rich-Harris:
you're welcome 😀

vesparny:
@kentcdodds 

I tried removing the html attributes list from glamorous.
The final build is 5.18kb gzipped (1kb less than actual one)

Obviously the only way to use glamorous will be, as you said:

```js
import glamorous from 'glamorous'

glamorous('a')({
  color: 'red'
})
```


Another possible solution would be

```js
function glamorous (comp, {rootEl, displayName} = {}) {
  .....
  ....
  // return glamorousComponentFactory
}

const handler = {
  get: (target, name) => glamorous.bind(null, name)
}

const proxiedGlamorous = new Proxy({}, handler)

export default proxiedGlamorous
```

this would enable users to do 

```js
import glamorous from 'glamorous'

glamorous.a({
  color: 'red'
})

// as well as

glamorous.whatever({
  color: 'red'
})
```

[...]

PR #69, opened 2017-04-19
feat(core): glamorous tiny

vesparny:
**What**:
This is a refactoring that makes it possible to use a `tiny` glamorous version.
A first attempt of realizing what discussed in #42 

**Why**:
Because sometimes size matters 

**How**:
by making `glamorous` configurable.

glamorous is now a factory function that accepts a `splitProps` function.
In the full fledged glamorous version, it will do as usual.
In the tiny version it will skip all the prop checks, forwarding everything to the underlying component.

It enables patterns like:

```jsx
import React from 'react'
import glamorous from 'glamorous/tiny'

const Comp = glamorous('div')({
  color: 'red'
}, (glam) => ({
  fontSize: glam.big ? 20 : 12
}))
const Root = () => (
  <Comp
    glam={{big: true}}
    thisWillBeForwardedAndreactWillWarn
     >
    ciao
  </Comp>
)

export default Root
```

and this is the gzipped size of glamorous tiny:

screen shot 2017-04-19 at 16 56 17

✨✨


kentcdodds:
NICE! I can't wait to give this a good review. I'm giving workshops today and tomorrow, but I'll try to get to it soon!

kentcdodds:
Before even looking at it, I'd really like a test (maybe in `dist-tests`?) that verifies `glamorous-tiny` works as expected. I don't want to worry about it suddenly breaking use cases in the future because we removed too much or something.

src/tiny.js
```diff
@@ -0,0 +1,15 @@
+import createGlamorous from './create-glamorous'
```
kentcdodds: This is remarkably simple!

vesparny:
Yep, I have created a basic test that checks `glam` prop is not passed down.

For sure we need to make proper tests, but I did not really change that much the current API.

I basically simplified the `render` method, making `splitProps` configurable.

I just want to check with you and other reviewers if we can make the `create-glamorous` factory a little bit better.

I am not sure what I did is the best...but it works.

enjoy your workshops 😉

vesparny:
Another thing we need to do is to change the rollup config in order to generate a standalone `tiny`, `withTheme` and `ThemeProvider`.

I know libraries like `recompose` and `lodash` do it, but I do not know how.

Probably this should be another PR tho.

kentcdodds:
Hoping to have @paulmolluzzo, @jackyho112, @kkemple and anyone else give this a review as well. I'll probably merge it tonight if I don't hear back.

src/constants.js
```diff
@@ -0,0 +1 @@
+export const CHANNEL = '__glamorous__'
```
paulmolluzzo: Nitpicking, but I get a lint warning here: `1:1  warning  Prefer default export  import/prefer-default-export`

paulmolluzzo:
👍 

> Another thing we need to do is to change the rollup config in order to generate a standalone `tiny`, `withTheme` and `ThemeProvider`.

This would be 💯 , mostly for a standalone `tiny`.

src/constants.js
vesparny: yep I see it as well, do you know how to tell eslint to skip this check?

paulmolluzzo: Option 1 is certainly the simplest:

1. Disable eslint for this line.
2. Change `constants.js` to this:

```js
const CHANNEL = '__glamorous__';
export default CHANNEL;
```
And then change all the `{CHANNEL}` references to just `CHANNEL`.

3. Update the `package.json` to have this for the `eslintConfig` block:

```js
"eslintConfig": {
    "extends": [
      "kentcdodds",
      "kentcdodds/jest",
      "kentcdodds/react",
      "kentcdodds/prettier"
    ],
    "rules": {
      "import/prefer-default-export": 0
    }
  },
```

That third option will affect a rule we might want to keep around. @kentcdodds?

kentcdodds: Yeah, let's go with the third option.... I expect we'll put more in here later, but I don't want this rule slowing us down now.

vesparny: Ok. I do not have access to my laptop now. Feel free to fix it yourself, otherwise will do it tomorrow

paulmolluzzo: Pushed the change since I already had it from when I wrote the comment.

src/create-glamorous.js
```diff
@@ -0,0 +1,162 @@
+/*
+ * This is a relatively small abstraction that's ripe for open sourcing.
+ * Documentation is in the README.md
+ */
```
kentcdodds: Can we actually remove the comment above? This was a comment that I wrote when glamorous was still in the PayPal codebase :smile:

kentcdodds:
I'm going to work on this a little bit more actually... Give me a bit.

kentcdodds:
Ok friends. If you checkout this commit: https://github.com/paypal/glamorous/pull/69/commits/0c4012a87a1ec78876fd7ffee81a01ed96587e96

Then you'll see that I've documented things and I made a few slight changes/additions:
1. Instead of passing `props.glam` we just pass all the props. I can see situations where someone would want all the props even if some of them are also forwarded, so It makes more sense to me that they're all passed whether you're using tiny or not.
2. I updated the rollup config to accommodate a `tiny` version for each of the builds that we currently have. I didn't really see a reason why not.
3. I updated the `dist-test` to include the tiny versions and also added a check for whether the import has the extra exports on it. I'm glad that we have this test!
4. I slightly updated the test to make sure it indicated what the dynamic props will be called with
5. I renamed a method
6. I plucked off more props in both implementations of `split-props`.

Let me know what you all think of these changes!

Next up, we should probably make builds for the `ThemeProvider` and `withTheme`.

src/create-glamorous.js
```diff
@@ -0,0 +1,162 @@
...
+          }
+        }
+
+        componentWillUnmount() {
+          // cleanup subscription
+          this.unsubscribe && this.unsubscribe()
+        }
+
+        render() {
+          // in this function, we're willing to sacrafice a little on
```
jackyho112: A small typo here. Is the performance gain referring to `const props = this.props`?

kentcdodds: It's referring generally to the function itself. Any codepath in this function may suffer a bit on readability for the sake of performance. That's all this comment is intended to say. How could it be improved?

jackyho112: Thanks for the clarification! I now realize this comment is more about describing what this function is aiming for. I was trying to find out/understand where exactly readability is sacrificed for performance 😄 .

I do think it may be useful to point those trade-offs out if it's not obvious because further contributors may refactor some code without realizing what it's good for performance or feel scared to touch the code because it may be done for performance. 

But either way, I think the way it currently is is fine. 👍

vesparny:
looks good to me

kentcdodds:
Awesome! Great work everyone! :clap:

PR #77, opened 2017-04-21
docs(logo): change the logo

kentcdodds:
ref #1

**What**: This changes the logo

**Why**: Because things discussed in #1 (copyright issues and such)

**How**: Help from friends like @Connorelsea and @markdalgleish :clap:


kentcdodds:
I would really like to have an SVG version of both of these. And actually, now that I think of it, I want to see a version of the full one with that gradient too 🤔

Connorelsea:
I'm not too familiar with editing SVGs, but the gradient is from #e63a5c to #f87880 if someone would like to try it out.

kentcdodds:
Could you try it on [this](https://codesandbox.io/s/D9O2NGE9B) @Connorelsea?

paulmolluzzo:
@kentcdodds Please tell me I can get this as a sticker! 🙏 🙏 🙏

kentcdodds:
YES!

Issue #63, opened 2017-04-17
concat forwardProps option for wrapped components


kentcdodds:
We should probably do the same to `forwardProps` that we do to `styles` [here](https://github.com/paypal/glamorous/blob/424a7e575f4242044b5c891209578f671e193de5/src/index.js#L138). Because otherwise you have to do this:

```javascript
const compProps = ['a', 'b']
const MyComp = glamorous(Comp, {forwardProps: compProps})()
const MyWrapperComp = glamorous(MyComp, {forwardProps: compProps})()
```

I would prefer to only have to do:

```javascript
const compProps = ['a', 'b']
const MyComp = glamorous(Comp, {forwardProps: compProps})()
const MyWrapperComp = glamorous(MyComp)()
```

And the `compProps` are forwarded along because the wrapping `MyComp` has defined them as acceptable props to be forwarded.

miguelc1221:
I'll give this a shot

miguelc1221:
I get an eslint complexity warning of 6 when adding the ternary to `forwardProps`, wondering if I should disable the eslint rule for this line?

kentcdodds:
If there's no way to break up that into a function, then yes, go ahead and use `// eslint-disable-next-line complexity` on the line above the function.

PR #80, opened 2017-04-22
refactor: concat forwardProps option

miguelc1221:
#63

**What**: concat forwardProps option metadata

**Why**:  Necessary in order to not have to pass in the same props again. See #63.

**How**: Created a function that returns a ternary in order to avoid the eslint complexity rule. Used the function on the `forwardProps` metadata in order to return correct forwardsProps.


miguelc1221:
I added unit but don't have much experience writing tests. Let me know if there is anything i should fix = ). Thanks

src/__tests__/index.test.js
```diff
@@ -105,6 +105,15 @@ test('merges composed component styles for reasonable overrides', () => {
   expect(wrapper).toMatchSnapshotWithGlamor()
 })
 
+test('merges composed component forwardProps', () => {
+  const parent = ({shouldRender, ...rest}) =>
+    (shouldRender ? <div {...rest} /> : null)
+  const Child = glamorous(parent)({forwardProps: ['shouldRender']})
```
kentcdodds: This should be:

```javascript
glamorous(parent, {forwardProps: ['shouldRender']})()
```

kentcdodds:
Thanks a ton for this work by the way. I'm excited to get this in!

src/__tests__/index.test.js
miguelc1221: ah you're right, fixed.

PR #84, opened 2017-04-24
docs(Examples): Example of using polished with glamorous

bhough:
Adds a couple examples of using polished mixins with glamorous as well as a sandbox that includes
polished examples.

Adding a couple examples of using `✨ polished` mixins with `glamorous` as well as a sandbox for users to play with using both together.

`✨ polished` is a general purpose library that aims to work with all `css-in-js` libraries. 

Additions to the examples README.md


kentcdodds:
Could we make the sandbox less noisy? Just make it a single file with basically the same stuff you have for the example. That way it's less confusing for people. Thanks!

bhough:
Yep, I will take a look at that later today.

kentcdodds:
Great! I'll go ahead and merge as-is. Just update the sandbox when you have time 👍

bhough:
Thanks @kentcdodds, sandbox is cleaned up now as well.

Issue #91, opened 2017-04-29
Built-in rect component doesn't render size/position attributes


AlteredConstants:
- `glamorous` version: 3.11.2
- `glamor` version: 2.20.24
- `react` version: 15.5.4

Relevant code.

```javascript
const BuiltinComponentRect = glamorous.Rect;
const BuiltinFactoryRect = glamorous.rect({ fill: "blue" });
const CustomRect = glamorous("rect", {
  forwardProps: ["x", "width", "height"]
})({ fill: "orange" });

ReactDOM.render(
  <div>
    <svg width="250" height="50">
      <rect x="0" width="50" height="50" fill="red" />
      <BuiltinComponentRect x="60" width="50" height="50" fill="green" />
      <BuiltinFactoryRect x="110" width="50" height="50" />
      <CustomRect x="170" width="50" height="50" />
    </svg>
  </div>,
  document.getElementById("root")
);
```

What you did:

N/A

What happened:

In Firefox and Edge, the green (built-in component) and blue (built-in factory-created component) boxes do not render correctly (they have no size) and in Chrome, the blue box doesn't render.

Reproduction:

https://codepen.io/alteredconstants/pen/YVZByv?editors=0010

Problem description:

I could very well be misunderstanding something, but it appears that the `x`, `width`, and `height` props are not being passed on to the underlying `<rect>` element with the built-in component/component factory. Inspecting the DOM in Chrome, it looks like `glamorous.Rect` is rendering those props as styles rather than attributes (which appears to be unsupported in Firefox) and the `glamorous.rect`-generated component simply doesn't render them at all.

Suggested solution:

I haven't used this library much (I like it so far!) but the docs suggest that non-style props on the built-in components should be passed through as attributes (e.g. `href` on `<a>`). I'm guessing these size/position ones aren't because they're also style properties? Considering this doesn't currently work on Firefox and Edge, I'd suggest adding those as exceptions for SVG components as they are effectively unusable otherwise.

Or since that might break things for people who only care about Chrome, maybe just a note about this in the docs recommending using the custom component syntax in this case.

kentcdodds:
Thanks! Probably a bug in our react-html-attributes dependency actually... On mobile now, but if you want to look into that, that'd be great!

jackyho112:
I believe it is. I will take a look!

AlteredConstants:
Yeah, it looks like that library only handles the `<svg>` element by assigning every property React mentions for all of the SVG components.

[...]

PR #93, opened 2017-04-30
Support SVG-specific attributes for all available SVG tags

jackyho112:
**What**:
This is an attempt to fix https://github.com/paypal/glamorous/issues/91. Currently, we only support SVG attributes on the `svg` element. This updates the library to support all of them.

**Why**:
React supports many SVG attributes that need to be forwarded to SVG elements other than `svg`.

**How**:
Check if the tag belongs to a SVG element and if so, check if it's one of the React-supported SVG attributes. 

Currently, this branch is not passing all the tests because some HTML elements share the same tag names with some SVG elements. So don't merge it obviously. Let's discuss what's the best way to deal with situations like that.

### DIFF
--- a/src/should-forward-property.js
+++ b/src/should-forward-property.js
@@ -30,7 +30,7 @@ const isCustomAttribute = RegExp.prototype.test.bind(
 
 const isSvgTag = tagName => supportedSVGTagNames.indexOf(tagName) !== -1
 const isHtmlProp = (name, tagName) => {
-  let elementAttributes
+  let elementAttributes = []
 
   if (isSvgTag(tagName)) {
     // all SVG attributes supported by React are grouped under 'svg'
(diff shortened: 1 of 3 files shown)


kentcdodds:
> Currently, this branch is not passing all the tests because some HTML elements share the same tag names with some SVG elements. So don't merge it obviously. Let's discuss what's the best way to deal with situations like that.

Could you write a test-case to highlight this specific issue? That will help me understand what you mean so we can talk about how to solve the problem :) Thanks!

jackyho112:
@kentcdodds nvm. I figured it out! Now the library should properly support all SVG attributes forwarding for all SVG elements supported by React.

src/should-forward-property.js
```diff
@@ -11,28 +11,40 @@ import reactHTMLAttributes from 'react-html-attributes'
...
+    // all SVG attributes supported by React are grouped under 'svg'
+    elementAttributes = reactHTMLAttributes.svg
+  } else {
+    elementAttributes = reactHTMLAttributes[tagName]
+  }
 
-  return globalReactHtmlProps.indexOf(name) !== -1 ||
+  return (
+    globalReactHtmlProps.indexOf(name) !== -1 ||
     (elementAttributes === undefined ?
```
kentcdodds: Maybe instead of this ternary, we could initialize `elementAttributes` to an empty array.

AlteredConstants:
Nice! Just pulled the latest and it works great for me. Awesome quick turnaround. Thanks!

jackyho112:
@kentcdodds thanks for helping out and the approval! 👍

PR #94, opened 2017-04-30
feat(typescript): improved css props typings

ErikCupal:
**What**:
There are lots of CSS rules that could have better typings. Therefore, I took [the interface `CSSProperties` from @types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts) (the interface that we use now) and did my best to improve it. Basically, I checked every prop in CSS reference on MDN and improved the prop type so that it matches its formal syntax.

Example:

```ts
type CSSWideKeyword = 'initial' | 'inherit' | 'unset'

...
// before
fontKerning?: CSSWideKeyword | any;

// Formal syntax on MDN: auto | normal | none
// https://developer.mozilla.org/en-US/docs/Web/CSS/font-kerning#Formal_syntax

// now
fontKerning?: CSSWideKeyword | 'auto' | 'normal' | 'none'
```

For many props it is not possible to get proper typechecking. Example is `margin` - the best what we can do now is type `string | number`. Hopefully, [this will improve in the future](https://github.com/Microsoft/TypeScript/issues/6579) 😊.

*Note:* It is necessary to include this interface directly in glamorous typings (instead of just updating the React typings themselves), so that we can get proper types for pseudoelements like `:hover` or `:focus`.

**What's missing (for this moment)**:
As I have written above, props that have dynamic typings (for example **margin**, **fontSize**, **bakground**...) cannot be checked properly now. I hope Typescript gets [support for this in the future](https://github.com/Microsoft/TypeScript/issues/6579).

For the same reason, there is not proper typechecking for `@media rules` yet.

**Why**:
Because better linting & typechecking means more fun and less bugs 😎.

**How**:
Manually 😀.

I'll be happy to see any comments or improvements 🙂.

### DIFF
--- a/typings/css-properties.d.ts
+++ b/typings/css-properties.d.ts
@@ -1,0 +1,1523 @@
+// See CSS 3 CSS-wide keywords https://www.w3.org/TR/css3-values/#common-keywords
+// See CSS 3 Explicit Defaulting https://www.w3.org/TR/css-cascade-3/#defaulting-keywords
+// "all CSS properties can accept these values"
+type CSSWideKeyword = 'initial' | 'inherit' | 'unset'
+
+// Based on [CSSProperties from React types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts)
+export interface CSSProperties {
+
+  /**
+   * Aligns a flex container's lines within the flex container when there is extra space in the cross-axis, similar to how justify-content aligns individual items within the main-axis.
+   */
+  alignContent?: CSSWideKeyword | 'flex-start' | 'flex-end' | 'center' | 'space-between' | 'space-around' | 'stretch'
+
+  /**
+   * Sets the default alignment in the cross axis for all of the flex container's items, including anonymous flex items, similarly to how justify-content aligns items along the main axis.
+   */
+  alignItems?: CSSWideKeyword | 'flex-start' | 'flex-end' | 'center' | 'baseline' | 'stretch'
+
+  /**
+   * Allows the default alignment to be overridden for individual flex items.
+   */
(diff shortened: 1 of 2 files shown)


ErikCupal:
@kentcdodds Sure 🙂. @luke-john do you want to look at it as well?

sammkj:
@ErikCupal I agree that some of the CSS typings are not up-to-date. But I feel like we shouldn't be writing this ourselves because we would need to constantly maintain it. Have you looked at https://github.com/typestyle/typestyle? Perhaps we can collaborate with them and use that instead?

luke-john:
This looks like an awesome addition.

As with @sammkj I'd be keen if this ended up in another repository for css typings which could be used by other libraries, however would be happy for this to land in the meantime as it will bring improvements today :).

Breaking it in a seperate repo down the track would also mean we could have extensive tests for it there and keep the glamorous ts tests to it's own functionality.

ErikCupal:
@sammkj @luke-john Yes, I absolutely agree we should have a repo for css typings. I know TypeStyle and used it before styled-components and glamorous. @basarat @notoriousb1t Do you guys think we could work on this together? There probably are/will be other css-in-js solutions that would benefit as well 😊. My idea is that there would be a common CSS properties interface that every css-in-js solution could extend for specific needs (like `$unique` or `$nest` in TypeStyle).

@sammkj @luke-john I also think we can approve this in the meantime to have better typings now ;).

notoriousb1t:
@ErikCupal I am open the idea of creating a decentralized CSS definition for TypeScript. However, the definition in this PR would be **incompatible** with our current types in TypeStyle.  TypeStyle uses a $nest property to avoid looking type-checking on nested selectors.  You can't directly nest selectors next to the other CSS properties.

https://github.com/typestyle/typestyle/issues/40

@basarat, what are your thoughts?

basarat:
@notoriousb1t I agree with you :rose: :heart:

PR #96, opened 2017-05-01
fix(typings): fix typings for wrapping stateless components

luke-john:
**What**:

Re adding the ability to wrap stateless components which was lost in a previous refactor

**Why**:

Otherwise typescript will error when wrapping a stateless component

**How**:

Added tests for wrapped components and then added typing to pass them :).


typings/styled-function.d.ts
```diff
@@ -4,7 +4,9 @@
  * @see {@link https://github.com/paypal/glamorous/blob/master/src/create-glamorous.js#L28-L131}
  */
 
-export type GlamorousComponent<P> = React.ComponentClass<P & ExtraGlamorousProps>
+export type GlamorousComponent<P> = 
+  | React.ComponentClass<P & ExtraGlamorousProps>
+  | React.StatelessComponent<P & ExtraGlamorousProps>
```
sammkj: `glamorousComponentFactory` returns a component class only. So it cannot be a union or else it would break HOC that wraps a Glamorous component. I think you meant to update https://github.com/paypal/glamorous/blob/master/typings/glamorous.d.ts#L25.

luke-john: Ahh yep, will update as follows.

```
export interface GlamorousInterface extends HTMLGlamorousInterface, SVGGlamorousInterface {
  <P>(
    component: 
      | React.ComponentClass<P>
      | React.StatelessComponent<P>,
```

PR #98, opened 2017-05-01
fix(typescript): fixed typings tests and typings for extended component

ErikCupal:
**What**:
There are two little fixes in this PR.

Firstly, I forgot to apply new CSSProperties typings (#94) to extended components (e.g. `const MyAwesomeLink = glamorous(MyLink)({ color: 'red' })`). It was still using the React ones. Fixed now. Sorry for that, overlooked it :).

Secondly, the tests were broken. Actually, the new typings revealed it (yey 😀). Fixed.

**Why**:
Because we want good typings

**How**:
Pretty easily! See the commits.

@luke-john @sammkj Would you review it :)?

PS: @luke-john I don't know why, but I can't you add as a reviewer, weird 😕.

### DIFF
--- a/test/glamorous.test.tsx
+++ b/test/glamorous.test.tsx
@@ -24,7 +24,7 @@ const Divider = glamorous.span(
     "fontSize": "10px",
     "zIndex": "auto"
   },
-  (props, theme: { main: { color: string; }}) => ({
+  (props, theme: { main: { color: string; } }) => ({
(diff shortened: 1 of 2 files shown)


luke-john:
Lgtm 👍. Thanks.

luke-john:
And I'd forgotten to accept the invitation for access. Have now accepted so you should be able to add me in the future ☺️

PR #101, opened 2017-05-04
docs(examples): add animation codesandbox

kentcdodds:


kentcdodds:
Feedback welcome as always.

Issue #89, opened 2017-04-27
require("glamorous").default undefined?


sammkj:
- `glamorous` version: 3.11.1
- `glamor` version: 2.20.4
- `react` version: 15.5.4

What you did:

I'm creating an UI component package written in Typescript.

What happened:

The UI component is written in ES6. Then Typescript compiles ES6 code to ES5. But it throws error when using the package.

Here's the compiled ES5 code:

![image](https://cloud.githubusercontent.com/assets/9153498/25500225/3aa7111a-2b5c-11e7-9388-d379337a97d1.png)

Problem description:

The error says that `glamorous.default` is `undefined`

kentcdodds:
Ah, that's interesting that TypeScript does that... 🤔 

So here's the deal with our various [dist files](https://unpkg.com/glamorous@3.11.1/dist/). The `main` in the `package.json` points to `dist/glamorous.cjs.js` which exports itself as a normal CJS module (so it doesn't have a `default`). With babel this isn't a problem because of the `interopRequireDefault` thing that it does. With webpack this isn't a problem, because it picks up the `module` config in our `package.json` which points to `dist/glamorous.es.js` and it can consume the ESModules. What do we need to do to make things work well with TypeScript?

sammkj:
I'm not sure what can be done. Either there's something wrong with the Typescript definition or Typescript itself. Looking through SO, I found [this](http://stackoverflow.com/questions/39109027/write-a-declaration-file-for-a-default-export-module), which is very similar to this case.

kentcdodds:
Feel free to open a PR to update whatever we need to so we can support TypeScript well!

dragosbulugean:
having the same issue with typescript. solution is to use require instead of import, but you lose autocomplete and it looks ugly next to all your imports.

kentcdodds:
Are you all using webpack?

ErikCupal:
I know I have been having problems with this some time ago (because of some different library). I solved it by compiling typescript to ES6 (with ES modules - no CJS), running through Babel and finally through Webpack (of course using ts-loader and babel-loader). I'm using it since them and got no problems 😃. In my opinion it is better solution for now than importing using require.

If anyone needs help with this, I can share my webpack.config.js, .babelrc and tsconfig.json 😉.

sammkj:
@dragosbulugean you're using `import glamorous = require("glamorous");`? I'm hesitant to do that because that does not seem like a valid ES6 import statement.

[...]

PR #104, opened 2017-05-10
fix(typescript): updated entry file for typescript consumption

luke-john:
**What**:

Fix for https://github.com/paypal/glamorous/issues/89

**Why**:

Rollup compilation results in glamorous unable to be used when compiled with typescript.

https://github.com/Microsoft/TypeScript/issues/13017#issuecomment-268657860


sammkj:
ahhh. so this is the cause! thanks @luke-john

Issue #108, opened 2017-05-17
Make CSSProperties interface exportable for TypeScript


bnwan:
Please let me know if below does not make sense :)

Problem description:
No exported member when doing this: 
`import glamorous, { CSSProperties } from 'glamorous';`

Suggested solution:
export `CCSProperties` in main glamorous definition file

I'll submit a PR :)

bnwan:
I've also added a few extra types to `propertyName` in `css-properties.d.ts`. 
From `[propertyName: string]: string | number | CSSProperties | undefined` to `[propertyName: string]: string | number | boolean | Function | CSSProperties | undefined`

I'll submit a PR very soon

Issue #81, opened 2017-04-23
Fix Typescript typings to work with dynamic styles


ErikCupal:
- `glamorous` version: 3.10.0
- `react` version: 15.4.2

Currently when we want to use dynamic styles in Typescript:
```tsx
const MyStyledDiv = glamorous.div(
  {
    margin: 1,
  },
  (props) => ({ // If you are using noImplicitAny rule you get an error:
                // Parameter 'props' implicitely has 'any' type
                // It can be fixed by simple supplying the type
                // like this `(props : { noPadding?: boolean }) => `
    padding: props.noPadding ? 0 : 4,
  })
)

<MyStyledDiv />
<MyStyledDiv noPadding /> // You get error here as well:
                          // Property 'noPadding' does not exist on type 'IntrinsicAttributes & HTMLProps<HTMLDivElement> & { children?: ReactNode; }'
```
We get these two errors 😕 (commented in the snippet).

It can be fixed pretty nicely. By making `glamorous.div` (and all other html element functions) accept optional generic with custom props we solve both errors and get proper intelliSense with all its glory 😎.

How does the code look with the optional generic:
```tsx
const MyStyledDiv = glamorous.div<{ noPadding?: boolean }>(
  {
    margin: 1,
  },
  (props) => ({
    padding: props.noPadding ? 0 : 4,
  })
)

<MyStyledDiv />
<MyStyledDiv noPadding />
```

It is just needed to make a few typings changes in [`typings/styled-function.d.ts`](https://github.com/paypal/glamorous/blob/master/typings/styled-function.d.ts):

```ts
export type StaticStyles<Properties> = Partial<Properties>
export type DynamicStyledFunction<Properties, CustomProps> = ( // added CustomProps generic
  props: CustomProps, // changed from `props?: object`
  theme?: object,
) => Partial<Properties>;

type Styles<Properties, CustomProps> = Array<DynamicStyledFunction<Properties, CustomProps> | StaticStyles<Properties>> // added CustomProps generic

...

export interface StyledFunction<Props, Properties> {
  <CustomProps>( // added CustomProps generic
    style1: StaticStyles<Properties>,
    ...styles: Styles<Properties, CustomProps> // added CustomProps generic
  ): GlamorousComponent<Props & CustomProps>; // added CustomProps generic
}

export interface StyledFunction<Props, Properties> {
  <CustomProps>( // added CustomProps generic
    style1: StaticStyles<Properties>,
    style2: StaticStyles<Properties>,
[...]

PR #111, opened 2017-05-18
Cssproperties

bnwan:
This PR is based on the issue below:
#108


typings/css-properties.d.ts
```diff
@@ -1525,5 +1525,5 @@ export interface CSSProperties {
   '::first-line'?: CSSProperties
   '::selection'?: CSSProperties
 
-  [propertyName: string]: string | number | CSSProperties | undefined
+  [propertyName: string]: string | number | boolean | Function | CSSProperties | undefined
```
ErikCupal: Could you provide an example where this would be useful 🙂? I'm not sure whether it is really needed

ErikCupal:
Thanks for PR 🙂. Yeah exporting the CSSProperties interface is not a bad idea, it might be useful for some usecases. It just isn't clear to me when would we need to assign `boolean` or `Function` to `[propertyName: string]`... Could you provide an example where this would be useful?

bnwan:
Use case is when you want to pass a custom prop that is either a `boolean` or a `Function`. Obviously this is not a problem when using ES6, but with TypeScript they are not recognised. An example is below:

```
const Item = glamorous.li({
    borderWidth: 5,
    borderStyle: 'solid',
    marginRight: 5,
    marginBottom: 5,
    padding: 10,
    cursor: 'pointer'
  },
    (props: CSSProperties) => ({
      borderColor: props.isDefault? colors.actionColor : colors.borderColor,
    })
  );

 <Item
    key={doc.optionId}
    flagDefault={doc.flagDefault}
    onClick={() => onClick(doc)}
  >  
</Item>
```

ErikCupal:
Aaah, I see.

There is no point in doing this. It can be done simply by generics. It should be in the docs, sorry for that. Here is an example with the generics:

```ts
const Item = glamorous.li<{ isDefault: boolean }>(
  {
    borderWidth: 5,
  },
  props => ({
    borderColor: props.isDefault ? 'red' : 'blue',
  })
)

const AnotherItem = () => (
  <Item isDefault={true}>
    Foo
  </Item >
)
```

This way you get 100% typechecking for custom dynamic properties. Look at this issue for more info #81
 🙂. 

Therefore the `boolean` and `Function` does not make any sense. Though I think that it can be useful to export the CSSProperties (e.g. when you need to create the object with styles alone and then pass its reference to glamorous).

bnwan:
Ahhh...that makes sense...much better. Thank you 😄 

What about if you need to do something like this:

```
<glamorous.Div
   css={style}
   onClick={() => this.onSave()}
 >
 </glamorous.Div>
```

TypeScript complains about `onClick`. Is there a better way to do this?

sammkj:
@bnwan the type definitions for built-in glamorous is incomplete, almost none I would say. To get rid of the `onClick` error, you need to add `React.HTMLAttributes<HTMLDivElement>` to `Div`.

ErikCupal:
@bnwan I think we can close this PR. Is it ok 🙂?

bnwan:
@ErikCupal Yes, but it'll be great if we can export `CSSProperties`

kentcdodds:
> @ErikCupal Yes, but it'll be great if we can export CSSProperties

Let's make that happen in another PR :+1:

bnwan:
@kentcdodds Great, I'll create another PR for it...thanks :)

ErikCupal:
@bnwan Yup, I agree it would be good. Just for now: you can import the CSSProperties using `import { CSSProperties } from 'glamorous/typings/css-properties'` ;), though it is better when it is exported directly

PR #122, opened 2017-05-22
docs($typescript): added typescript usage notes and known issues

luke-john:
closes #83 

**What**:
Added usage notes and known issues for typescript

**Why**:
The current bundled typescript definitions are incomplete and based around the needs of the developers who contributed them.

Whilst this is true of the definitions of many non typescript projects, it could still be a pain point for consumers.

**How**:
Based on feedback of #83 

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [x] Followed the commit message format

PR #124, opened 2017-05-23
fix($typescript): Add ExtraGlamorousProps type as top-level export

kryops:
**What**:
This adds the `ExtraGlamorousProps` type as a top-level export and adds a test for the underlying TypeScript issue it works around.

**Why**:
This fixes issues when using TypeScript with `"declaration": true` in combination with webpack.

Original TypeScript issue: https://github.com/Microsoft/TypeScript/issues/5938

> [ts] Exported variable 'StyledPromiseCard' has or is using name 'ExtraGlamorousProps' from external module "[...]/node_modules/glamorous/typings/styled-function" but cannot be named

One workaround of this issue is to explicitly import the misbehaving type:
```ts
import { ExtraGlamorousProps } from 'glamorous/typings/styled-function'
```
This does however break the module resolution in our webpack build for Node.JS and throws "Module not found" errors.
With this PR, the type can now be imported directly from "glamorous":

```ts
import glamorous, { ExtraGlamorousProps } from 'glamorous'
```

**How**:
To make the test work completely, the TypeScript test had to actually emit files, otherwise the error case could not be reproduced. Added `rimraf` as an explicit dependency to clean up after the test.

**Checklist**:

- [ ] Documentation should be added to https://github.com/paypal/glamorous/pull/122
- [x] Tests
- [x] Code complete

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@ coverage
 dist
 .opt-in
 .opt-out
+test-ts
 typings/*.js
 *.log
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "start": "nps",
     "test": "nps test",
-    "test:ts": "tsc -p ./tsconfig.json",
+    "test:ts": "(tsc -p ./tsconfig.json && rimraf test-ts) || rimraf test-ts",
     "commitmsg": "opt --in commit-msg --exec \"validate-commit-msg\"",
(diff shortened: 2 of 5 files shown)


kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)? Feel free to do so in another PR :)

kryops:
I will if I end up making more changes :wink:
Thanks for the super quick landing!

Issue #125, opened 2017-05-23
Support context


kentcdodds:
It'd be useful to be able to do:

```javascript
const MyDiv = glamorous.div((props, theme, context) => ({ /* my styles */ }))
MyDiv.contextTypes = { /* context types */ }
```

I've not really used context a ton, so I'd appreciate feedback on this API before anyone start working on it. cc @kenwheeler

kentcdodds:
Potentially could do:

```javascript
const MyDiv = glamorous.div((props, theme) => ({ /* props.context has context? Maybe kinda confusing... */ }))
MyDiv.contextTypes = { /* context types */ }
```

kentcdodds:
What if we made a breaking change. So folks using theme today like:

```javascript
const MyDiv = glamorous.div((props, theme) => ({ /* my styles */ }))
```

Would have to do:

```javascript
const MyDiv = glamorous.div((props, {theme}) => ({ /* my styles */ }))
```

The difference being that the second argument is the `context` object which you're destructuring. I think that's acceptable. We could even create a codemod for that easily 👍

ajwhite:
I like the the original proposal of the 3 named arguments

```js
glamorous.div((props, theme, context) => ({})
```
However it won't scale nicely if more options are introduced later, and we have this leap-frog effect where we go from 
```js
(objFromReact, objFromGlamorous, objFromReact, objFromGlamorous, ...) => ...
```
And then the interface feels a bit fuzzy.

➡️ I like the idea of having the second argument be an object of "other options" that deviate from the most common denominator of working with props -- themes, contexts, and anything else down the pipeline. 👍

vesparny:
What would be a use-case for this feature?
I always thought about context as a thing to use for theming and stuff like that.

If someone really needs context within a glamorous component I think it's possible tu require context from parent and send down to glamorous comp via props.

kentcdodds:
For context (no pun intended), this issue was inspired by [this thread](https://twitter.com/kentcdodds/status/867034742973231104).

I like the idea of enabling glamorous components to do everything that regular function components can do (they can already do `propTypes` and `defaultProps` so it's a shame that you cannot do `contextTypes` too). One use case would be if you want to style things differently based on whether it matches the react-router route for example. I wouldn't want to make a HoC just for that. :)

kentcdodds:
[...]

PR #127, opened 2017-05-23
feat(context): add dynamic style context support

ajwhite:
**What**: `context` is now accessible in dynamic style functions

```js
const MyDiv = glamorous.div(
  (props, context) => ({ /* style from context */ })
)
MyDiv.contextTypes = { /* user context types */ }
```

**Why**: See goals in #125, previously `context` was inaccessible by `glamorous` dynamic styles

**How**: The `GlamorousComponent` now [passes `context` to `getGlamorClassName`](https://github.com/robinpowered/glamorous/blob/c96f7aaf30b2a7d7910d6cb44f9ba217397f9fd5/src/create-glamorous.js#L103) instead of `theme`, which [passes the `context` along to the dynamic style function](https://github.com/robinpowered/glamorous/blob/c96f7aaf30b2a7d7910d6cb44f9ba217397f9fd5/src/get-glamor-classname.js#L35).

The `theme` is still in the picture: we [add it to the context](https://github.com/robinpowered/glamorous/blob/c96f7aaf30b2a7d7910d6cb44f9ba217397f9fd5/src/create-glamorous.js#L96) only if the `context` does not already contain a `theme` (this would happen if the user has a `theme` in their `context`).

## Feedback requested prior to merging

The user is now able to provide their own `contextTypes`:

```js
const MyDiv = glamorous.div(
  (props, context) => ({ /* style from context */ })
)
MyDiv.contextTypes = { /* user context types */ }
// oh no! they overwrote the contextTypes that we wrote!
```

We still ultimately want `CHANNEL` to be defined on the `GlamorousComponent`'s `contextTypes` definition. 

[Here's the current solution](https://github.com/robinpowered/glamorous/blob/c96f7aaf30b2a7d7910d6cb44f9ba217397f9fd5/src/create-glamorous.js#L122-L146):
I have `Object.defineProperty` defining a custom `get` and `set` rule on `GlamorousComponent.contextTypes`, such that we can always append our `CHANNEL` onto the `contextTypes`. **We don't want to lose our original context definition.**

When the user assignes `TheirGlamorousComponent.contextTypes = ...`, it [updates a private local variable `userDefinedContextTypes`](https://github.com/robinpowered/glamorous/blob/c96f7aaf30b2a7d7910d6cb44f9ba217397f9fd5/src/create-glamorous.js#L132-L134).

When React reads the component's `contextTypes`, [the custom `getter` returns a merged object of the user's `contextTypes` and glamorous'' `contextTypes`](https://github.com/robinpowered/glamorous/blob/c96f7aaf30b2a7d7910d6cb44f9ba217397f9fd5/src/create-glamorous.js#L135-L145).

**This is has a bit of magic to it 🎩 🐰 , anyone have thoughts for a more straightforward approach?**

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete (**pending feedback**)

### DIFF
--- a/src/__tests__/index.js
+++ b/src/__tests__/index.js
@@ -5,6 +5,7 @@ import {render, mount} from 'enzyme'
 import * as jestGlamorReact from 'jest-glamor-react'
 import {oneLine} from 'common-tags'
 import glamorous from '../'
+import {PropTypes} from '../react-compat'
 
 import {CHANNEL} from '../constants'
 
@@ -278,7 +279,7 @@ test('renders a component with theme properties', () => {
     {
       color: 'red',
     },
-    (props, theme) => ({padding: theme.padding}),
+    (props, {theme}) => ({padding: theme.padding}),
   )
   expect(
     render(<Comp theme={{padding: '10px'}} />),
@@ -291,7 +292,7 @@ test('in development mode the theme is frozen and cannot be changed', () => {
     {
       color: 'red',
     },
-    (props, theme) => {
+    (props, {theme}) => {
       expect(() => {
         theme.foo = 'bar'
       }).toThrow()
@@ -309,7 +310,7 @@ test('in production mode the theme is not frozen and can be changed', () => {
     {
       color: 'red',
     },
-    (props, theme) => {
+    (props, {theme}) => {
       expect(() => {
         theme.foo = 'bar'
       }).not.toThrow()
@@ -325,7 +326,7 @@ test('passes an updated theme when theme prop changes', () => {
     {
       color: 'red',
     },
-    (props, theme) => ({padding: theme.padding}),
+    (props, {theme}) => ({padding: theme.padding}),
   )
   const wrapper = mount(<Comp theme={{padding: 10}} />)
   expect(wrapper).toMatchSnapshotWithGlamor(`with theme prop of padding 10px`)
@@ -403,4 +404,19 @@ test('can accept classNames instead of style objects', () => {
     'extra-thing',
   )
   expect(render(<Comp />)).toMatchSnapshotWithGlamor()
+})
+
+test('should accept user defined contextTypes', () => {
+  const Child = glamorous.div((props, context) => ({
+    fontSize: context.fontSize,
+  }))
+  Child.contextTypes = {
+    fontSize: PropTypes.number,
+  }
+
+  const context = {
+    fontSize: 24,
+  }
+
+  expect(render(<Child />, {context})).toMatchSnapshotWithGlamor()
 })
--- a/src/__tests__/theme-provider.js
+++ b/src/__tests__/theme-provider.js
@@ -5,6 +5,7 @@ import * as jestGlamorReact from 'jest-glamor-react'
 import glamorous from '../'
 import ThemeProvider from '../theme-provider'
 import {CHANNEL} from '../constants'
+import {PropTypes} from '../react-compat'
 
 expect.extend(jestGlamorReact.matcher)
 expect.addSnapshotSerializer(jestGlamorReact.serializer)
@@ -22,7 +23,7 @@ test('renders a component with theme', () => {
     {
       color: 'red',
     },
-    (props, theme) => ({padding: theme.padding}),
+    (props, {theme}) => ({padding: theme.padding}),
   )
   expect(
     render(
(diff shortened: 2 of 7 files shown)


src/__tests__/index.js
```diff
@@ -404,3 +405,29 @@ test('can accept classNames instead of style objects', () => {
...
+    render() {
+      return <Child />
+    }
+  }
+
+  Parent.childContextTypes = {
+    fontSize: PropTypes.number,
+  }
+
+  expect(render(<Parent />)).toMatchSnapshotWithGlamor()
```
kentcdodds: Thanks to [`enzyme`](https://github.com/airbnb/enzyme/blob/master/docs/api/render.md) we can actually provide the context without having a `parent`. So this should be able to work as:

```jsx
expect(render(<Childe />, {context: {fontSize: 24}}).toMatchSnapshotWithGlamor()
```

src/__tests__/theme-provider.js
```diff
@@ -60,11 +61,11 @@ test('theme properties updates get propagated down the tree', () => {
 })
 
 test('merges nested themes', () => {
-  const One = glamorous.div({}, (props, {padding, margin}) => ({
+  const One = glamorous.div({}, (props, {theme: {padding, margin}}) => ({
     padding,
     margin,
   }))
-  const Two = glamorous.div({}, (props, {padding, margin}) => ({
+  const Two = glamorous.div({}, (props, {theme: {padding, margin}}) => ({
```
kentcdodds: Oh rats, my codemod doesn't handle destructuring `theme` 😅

src/__tests__/theme-provider.js
```diff
@@ -109,3 +110,69 @@ test('does nothing when receive same theme via props', () => {
...
+
+test('does not override user defined theme contexts', () => {
+  const Child = glamorous.div((props, context) => ({
+    fontSize: context.theme.fontSize,
+  }))
+  Child.contextTypes = {
+    theme: PropTypes.object,
+  }
+
+  class Parent extends React.Component {
```
kentcdodds: Should be able to do the cool context thing with enzyme here too :)

src/create-glamorous.js
```diff
@@ -112,10 +119,33 @@ export default function createGlamorous(splitProps) {
...
       }
 
+      let userDefinedContextTypes = null
+
+      // configure the contextTypes to be settable by the user,
+      // however also retaining the glamorous channel.
+      Object.defineProperty(GlamorousComponent, 'contextTypes', {
+        enumerable: true,
+        configurable: true,
+        set: value => {
```
kentcdodds: nit-pick (feel free to ignore): I'd prefer: `set(value) {` and `get() {`

src/create-glamorous.js
```diff
@@ -112,10 +119,33 @@ export default function createGlamorous(splitProps) {
...
       }
 
-      GlamorousComponent.contextTypes = {
+      const defaultContextTypes = {
         [CHANNEL]: PropTypes.object,
       }
 
+      let userDefinedContextTypes = null
+
+      // configure the contextTypes to be settable by the user,
```
kentcdodds: These comments are helpful :+1:

src/create-glamorous.js
ajwhite: Oh me too, good catch

src/__tests__/index.js
ajwhite: Oh nice! Super helpful 🙌

src/__tests__/theme-provider.js
ajwhite: Ran into a little bit of trouble when injecting `context` with the `ThemeProvider`, who also defines a child context:

https://gist.github.com/ajwhite/96955b52c1f29962b4d5f1cf13ed6202

I'll do some research on this, unless you've ran into this and have some tips off the top of your head

kentcdodds: Ok, don't worry about it. Not worth the effort :) It's good as-is :)

ajwhite: Seems like I might need the `Parent` here to at least define the `childContextTypes`

ajwhite: Ah - now it wants `getChildContextTypes` 😅 

Bit of a rabbit hold - I'll keep the `ThemeProvider` ones as-is, but the test in `index.js` works great with the `enzyme` `context` injection.

ajwhite:
921ed59 amended by ac1bca2 with [enzyme's `context` injection](https://github.com/paypal/glamorous/pull/127#discussion_r118080491) and using the [object method syntax](https://github.com/paypal/glamorous/pull/127#discussion_r118081496)

kentcdodds:
Merging this into the `next` branch because it's breaking. Let's think of what else we could break if we need to...

PR #128, opened 2017-05-23
release context changes

kentcdodds:
**What**: This is a branch to group breaking changes

**Why**: So we can limit how many major releases we push

**How**: Merged a breaking change into `next`.

What else could we do as part of this major release? I'll wait about 24 hours before merging/releasing this.

PR #130, opened 2017-05-24
feat(context): move context to a third parameter

kentcdodds:
**What**: Changing the implementation of #128

**Why**: As discussed in #125

**How**:
- Remove/change unnecessary tests
- Update the implementation to pass both theme and context.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] ~~Added myself to contributors table~~ 


kentcdodds:
I'll go ahead and merge this because it's going into `next`. That way we can see more easily what will go into `master` and review _that_ :)

PR #131, opened 2017-05-24
feat: allow importing built-in glamorous components

### LINKED ISSUE #123: Allow importing built-in glamorous components
I think it would be awesome to allow something like ```import { Div, Span } from 'glamorous'```. It's not a huge deal to just pull it off below, but I get questions about this and was wondering if there is a possible solution.

It seems that we will need some type of [pre-build](https://gist.github.com/FezVrasta/a29f3d8f6a715993a05ba1d24a241d3b) or babel-transform to accomplish this, I'm not really familiar in this area, but I'm willing to help out wherever I might be able to.

marzelin:
**What**:
allow something like `import { Div, Span } from 'glamorous'`

**Why**:
Convenient helpers improve user experience.

**How**:
Create a build script that appends exports to the es file.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

Closes #123

Sidenote:
This is not related directly to this commit, but my liter shows an error in `dist/glamorous.es.js`: 
'A module cannot have multiple default exports.'
That's caused by line 1177: `exports.default = reactHtmlAttributes$2;`


kentcdodds:
Hi! Thanks so much for this! I'm going to pull it down and run it to double check, but it looks pretty great.

Here are the linting failures. You may want to run `npm start validate` to check things before you push your changes :)

screen shot 2017-05-24 at 11 39 55 am

kentcdodds:
I would love it if we could also get a test in the `dist-tests` to check whether you can import things as we want to. Can do it with just one or two elements I think. Also documentation updates would be useful.

kentcdodds:
As for your linter issues, I think your linter is busted unless I'm missing something. That file is fine I think.

other/concat-exports.js
```diff
@@ -0,0 +1,26 @@
...
+const dashToCamelCase = s =>
+  s.replace(/-([a-z])/g, ([, char]) => char.toUpperCase())
+
+const elementExports = htmlTagNames
+  .concat(svgTagNames)
+  .filter((tag, index, array) => array.indexOf(tag) === index)
+  .map(tag => {
+    const validJsName = dashToCamelCase(tag)
+    const capitalName = capitalize(validJsName)
+    return `export const ${capitalName} = glamorous['${tag}']`
```
kentcdodds: Let's add a semicolon here because this is our transpiled code, we should never rely on ASI.

marzelin:
- [x] fix bug where component names collide with JavaScript constructors
- [x] tests added
- [x] documentation added

other/concat-exports.js
```diff
@@ -0,0 +1,88 @@
...
+
+function capitalize(s) {
+  return s.slice(0, 1).toUpperCase() + s.slice(1)
+}
+
+function dashToCamelCase(s) {
+  return s.replace(/-([a-z])/g, ([, char]) => char.toUpperCase())
+}
+
+const jsBuiltInObjectNames = [
```
kentcdodds: I wonder whether there's an npm package (probably by Sindre Sorhus) that has these so we don't have to hard-code them 🤔

other/concat-exports.js
```diff
@@ -0,0 +1,88 @@
...
+  'Uint32Array',
+  'Uint8Array',
+  'Uint8ClampedArray',
+  'WeakMap',
+  'WeakSet',
+  'WebAssembly',
+]
+
+function unCollide(name) {
+  const prefix = 'html'
```
kentcdodds: Very interesting! Makes total sense. Could you explain why (perhaps in a code comment) the prefix `html` is used? Seems like a good one, just curious to know why.

kentcdodds: Oh, also, this should probably be capitalized. All the others have the first letter capitalized so it's:

```javascript
import {HtmlObject, Article} from 'glamorous'
```

Rather than:

```javascript
import {htmlObject, Article} from 'glamorous'
```

other/concat-exports.js
souporserious: Found one :) https://github.com/jonschlinkert/reserved

other/concat-exports.js
```diff
@@ -0,0 +1,88 @@
+const appendToFile = require('fs').appendFile
```
souporserious: Are you getting a warning with this? I had to switch to `appendFileSync` to get rid of it.

other/concat-exports.js
marzelin: the `reserved` package wasn't updated in 3 years and is a bit outdated. I can use node.js to generate the list.

other/concat-exports.js
marzelin: I chose `html` because users may already be familiar with this prefix since it is used in JavaScript (`htmlFor`). But since component names have to be capitalized (otherwise React/JSX treats the tag as custom element), I think postfix `Tag` (ie `ObjectTag`) is a nicer alternative. What do you think?

other/concat-exports.js
marzelin: I'm getting no warnings. I think async execution is not super useful here, so I can switch to sync.

marzelin:
- [x] change `html` prefix to `Tag` postfix when name collisions
- [x] automatically create a constructor names list
- [x] change `appendFile` to `appendFileSync`

README.md
```diff
@@ -131,6 +131,8 @@ If you're transpiling (and/or using the `jsnext:main`):
 
 ```javascript
 import glamorous, {ThemeProvider} from 'glamorous'
+// you can also import specific components
+import {Div, H2} from 'glamorous'
```
kentcdodds: Could you include a line about the special case for JavaScript built-ins like `Object`? Something like:

> // Tags with the same name as built-in JavaScript objects are importable with a `Tag` suffix, IE: `ObjectTag`.

other/concat-exports.js
```diff
@@ -0,0 +1,43 @@
...
+  .filter((tag, index, array) => array.indexOf(tag) === index)
+  .map(tag => {
+    const validJsName = dashToCamelCase(tag)
+    const capitalName = capitalize(validJsName)
+    // add postfix if name collides with js constructors
+    const tagName = unCollide(capitalName)
+    return `export const ${tagName} = glamorous['${tag}'];`
+  }).join`\n`
+
+const toWrite = `\n${elementExports}`
```
kentcdodds: It would be handy to add a comment for the first line that says:

> // these exports below are generated and will be tree-shaken if you're using Webpack 2 or Rollup

marzelin:
- [x] add a comment preceding exports
- [x] provide info about `Tag` suffix in documentation

other/concat-exports.js
```diff
@@ -0,0 +1,45 @@
...
+function dashToCamelCase(s) {
+  return s.replace(/-([a-z])/g, ([, char]) => char.toUpperCase())
+}
+
+function isCapitalized(s) {
+  return !!s && s[0] === s[0].toUpperCase()
+}
+
+// node version <= 7 does not support trailing commas in function calls
+// prettier-ignore
```
kentcdodds: TIL :)

other/concat-exports.js
```diff
@@ -0,0 +1,45 @@
...
+
+function dashToCamelCase(s) {
+  return s.replace(/-([a-z])/g, ([, char]) => char.toUpperCase())
+}
+
+function isCapitalized(s) {
+  return !!s && s[0] === s[0].toUpperCase()
+}
+
+// node version <= 7 does not support trailing commas in function calls
```
kentcdodds: Normally, I'll add [this](https://github.com/paypal/glamorous/blob/b7424288431878bace8d9bb3b7d641e21a61b78f/package-scripts.js#L111-L123) at the end of files which are not transpiled:

```javascript
// this is not transpiled
/*
  eslint
  max-len: 0,
  comma-dangle: [
    2,
    {
      arrays: 'always-multiline',
      objects: 'always-multiline',
      functions: 'never'
    }
  ]
 */
```

other/concat-exports.js
```diff
@@ -0,0 +1,45 @@
...
+// prettier-ignore
+const jsBuiltInConstructorNames = Object.getOwnPropertyNames(global)
+  .filter(isCapitalized)
```
kentcdodds: I'm unsure whether this is really necessary 🤔

other/concat-exports.js
```diff
@@ -0,0 +1,45 @@
...
+
+const componentExports = htmlTagNames
+  .concat(svgTagNames)
+  .filter((tag, index, array) => array.indexOf(tag) === index)
+  .map(tag => {
+    const validJsName = dashToCamelCase(tag)
+    const capitalName = capitalize(validJsName)
+    // add postfix if name collides with js constructors
+    const tagName = unCollide(capitalName)
+    return `export const ${tagName} = glamorous['${tag}'];`
```
kentcdodds: Ah! Close call, this should actually be:

```javascript
return `export const ${tagName} = glamorous['${capitalName}'];`
```

README.md
```diff
@@ -131,6 +131,12 @@ If you're transpiling (and/or using the `jsnext:main`):
 
 ```javascript
 import glamorous, {ThemeProvider} from 'glamorous'
+
+// you can also import specific components
```
kentcdodds: We should probably call out the fact that these are the built-in components by saying:

> Please read the section below on "Built-in" components.

other/concat-exports.js
marzelin: It's not necessary but reduces the number of items to compare from 63 to 40 which improves performance a bit.

other/concat-exports.js
marzelin: I was wondering what to put there. Isn't `glamorous.div` equal to `glamourous.Div`?

kentcdodds: It is not. Here's some terminology that should hopefully be helpful:

1. `glamorous` is a "componentFactoryFactory"
2. `glamorous.div` is a "componentFactory" which is created by calling `glamorous('div')`
3. `glamorous.Div` is a "component" which is created by calling (effectively) `glamorous('div')()` (in addition to a few extra things)

We don't (and wont?) support exporting the component factories, but exporting the components themselves could be useful.

I hope that helps clarify :)

other/concat-exports.js
kentcdodds: I'm more concerned about readability/maintainability of this file than performance and would prefer to not include anything in here that's only here for performance reasons. The script is really fast as it is :)

marzelin: Well, I think this code is quite readable but of course less code is always a plus. I'll remove it.

other/concat-exports.js
marzelin: I should've RTFM :blush: sorry

other/concat-exports.js
kentcdodds: Yeah, it's not about how much code, it's about the fact that you need to decide why this matters, and the fact is that it doesn't ;-)

Thanks so much!

marzelin: We're discussing it so it kind of matters (or it's just bikeshedding) ;)

other/concat-exports.js
marzelin: Why don't we just remove `componentFactory` and move its competencies onto a glamorous component? That would make things simpler.

kentcdodds: I've considered this, and I'm not totally against it. I think that it could make things even more complex from both an API and implementation standpoint, but I'm open to talk about it in another issue if you'd like to bring it up.

marzelin:
- [x] export `GlamorousComponents` instead of `GlamorousComponentFactories`
- [x] remove unnecessary function
- [x] improve documentation
- [x] make tests more robust

README.md
```diff
@@ -131,6 +131,13 @@ If you're transpiling (and/or using the `jsnext:main`):
 
 ```javascript
 import glamorous, {ThemeProvider} from 'glamorous'
+
+// you can also import specific Glamorous Components (see section below on "Built-in" components)
+import {Div, H2} from 'glamorous'
+
+// tags with the same name as built-in JavaScript objects are importable with a Tag suffix
+// and tag names that contain dashes are transformed to CamelCase
+import {MapTag, ColorProfile} from 'glamorous'
```
kentcdodds: This is perfect!

kentcdodds:
Wonderful! Thank you!

PR #132, opened 2017-05-25
Add README.md "Table of Contents"

PeterDaveHello:
This PR is going to add "Table of Contents" to README document to improve its readability so that users don't need to scroll down and up to find out where's the section they need.


ajwhite:
This seems super helpful, it is quite an extensive README. Thanks for adding this!

WDYT about 3 levels vs 4 levels?

|3 levels | 4 levels|
|----|----|
|screen shot 2017-05-24 at 8 30 31 pm | screen shot 2017-05-24 at 8 30 43 pm|

With 4 levels it feels pretty vertically tall. When viewing the README, the only content "above the fold" is the title, badges, and table of contents. But it looks like we have the same outcome with 3 levels.. Hrmm. The titles in the 4th level might also be helpful in identifying terminology ("using `className` might be a helpful waypoint).

paulmolluzzo:
Thanks @PeterDaveHello this is super helpful!

> But it looks like we have the same outcome with 3 levels, although it is 7 lines less.. Hrmm. The titles in the 4th level might also be helpful in identifying terminology ("using className might be a helpful waypoint).

@ajwhite makes a fair point.

I think that 3 levels is still enough information to glance over and find something I might need. I imagine that people new to `glamorous` are more likely to look for "Overriding component styles" than a specific subtopic of that, such as "using `glamorous()` composition".

Personally, I usually use a ToC to browse an unfamiliar topic and `Ctrl-f` when I'm looking for something specific. :man_shrugging:

kentcdodds:
This is super awesome! Thank you for making it! I would really prefer to not maintain this TOC though. Could we have it automated? A quick search showed [doctoc](https://www.npmjs.com/package/doctoc) as something we could use. I like it because it allows you to specify where the TOC goes and if you want to skip a heading.

We could hook it into [the `lint-staged` config](https://github.com/paypal/glamorous/blob/master/package.json#L99-L104) to run the CLI and `git add` for `README.md`. What do you all think?

ajwhite:
Oh wow, `doctoc` looks awesome, of course there's a tool for this 😂 

And look at that.. [they event have a setting for a max depth](https://www.npmjs.com/package/doctoc#specifying-a-maximum-heading-level-for-toc-entries) as discussed above.

An automated TOC sounds 👌  very neat.

PeterDaveHello:
Done.

kwelch:
I would have expected an edit to the `lint-staged` config. Am I missing something?

package.json
```diff
@@ -11,7 +11,8 @@
     "test": "nps test",
     "test:ts": "(tsc -p ./tsconfig.json && rimraf test-ts) || rimraf test-ts",
     "commitmsg": "opt --in commit-msg --exec \"validate-commit-msg\"",
-    "precommit": "lint-staged && opt --in pre-commit --exec \"npm start validate\""
+    "precommit": "lint-staged && opt --in pre-commit --exec \"npm start validate\"",
+    "doctoc": "doctoc --maxlevel 3 --title '**Table of Contents**' README.md"
```
kentcdodds: Super! In this project, we use `nps` which puts scripts in `package-scripts.js` Could you move this script to there? Then you'll run `npm start doctoc`. That way we can add a helpful `description` :+1:

PeterDaveHello: Any more specific reference? Not sure how would you want it to be, I thought https://github.com/paypal/glamorous/blob/b7424288431878bace8d9bb3b7d641e21a61b78f/package.json#L99-L104 would be good already?

kentcdodds: Sorry, basically, you can do the same thing as we're doing with the `lint` script [here](https://github.com/paypal/glamorous/blob/b7424288431878bace8d9bb3b7d641e21a61b78f/package-scripts.js#L70-L73). So you can add:

```javascript
    doctoc: {
      description: 'generates a Table of Contents in the README',
      script: 'doctoc --maxlevel 3 --title '**Table of Contents**' README.md',
    },
```

Then you can run that script with: `npm start doctoc` :+1:

package.json
```diff
@@ -8,6 +8,7 @@
   "typings": "typings/glamorous.d.ts",
   "scripts": {
     "start": "nps",
+    "doctoc": "doctoc --maxlevel 3 --title \"**Table of Contents**\"",
```
kentcdodds: Sorry to make you do all that `package-scripts.js` work only to move it back @PeterDaveHello. Things weren't working too well with the `nps` + `lint-staged` integration. I'll be working on that :)

kentcdodds:
@PeterDaveHello would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

PeterDaveHello:
Maybe just add that when you need to update readme as that's not so important?

PR #138, opened 2017-05-25
feat(className): add displayName to the className

### LINKED ISSUE #137: Add component name to the class name
In [the styled-components v2 blogpost announcement](https://medium.com/styled-components/announcing-v2-f01ef3766ac2) it says:

> To circumvent this limitation we’ve built a Babel plugin which allows us to generate classes that include your components’ names! While we were at it we also made sure the correct names show up in the React DevTools, so rather than seeing <styled.div> or <styled(Component)> you’ll now see <Wrapper> and <styled(Link)> !

We already have [a babel plugin](https://github.com/bernard-lin/babel-plugin-glamorous-displayname) to make the display name be more helpful (thank you @bernard-lin), but the className does _not_ include the component name and that would be really helpful.

I just had an idea as I was typing this that would be really easy... What if we take the `displayName` value and attach it as part of the className. I don't think that's too bad of a solution now that I think of it 🤔 

Hmm... I'm just going to make a PR for this... Would love some feedback on that. Stand by :)

kentcdodds:
**What**: Closes #137

**Why**: This makes it easier to know which elements map to which glamorous components.

**How**:
- Change the `className` to include the `GlamorousComponent.className`

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/src/__tests__/__snapshots__/index.js.snap
+++ b/src/__tests__/__snapshots__/index.js.snap
@@ -39,6 +39,12 @@ exports[`can accept classNames instead of style objects 1`] = `
 />
 `;
 
+exports[`can be configured to use the displayName in the className 1`] = `
+<div
+  class="css-nil Hi-There"
+/>
+`;
+
 exports[`can use pre-glamorous components with css attributes 1`] = `
 .css-1t62idy,
 [data-css-1t62idy] {
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -7,22 +7,26 @@ import {PropTypes} from './react-compat'
 import {CHANNEL} from './constants'
 import getGlamorClassName from './get-glamor-classname'
 
-export default function createGlamorous(splitProps) {
+export default createGlamorous
+
+function createGlamorous(splitProps) {
+  // TODO: in a breaking version, make this default to true
+  glamorous.config = {useDisplayNameInClassName: false}
+
+  return glamorous
+
   /**
-   * This is the main export and the function that people
-   * interact with most directly.
-   *
-   * It accepts a component which can be a string or
-   * a React Component and returns
-   * a "glamorousComponentFactory"
-   * @param {String|ReactComponent} comp the component to render
-   * @param {Object} options helpful info for the GlamorousComponents
-   * @return {Function} the glamorousComponentFactory
-   */
-  return function glamorous(
-    comp,
-    {rootEl, displayName, forwardProps = []} = {},
-  ) {
+  * This is the main export and the function that people
+  * interact with most directly.
+  *
+  * It accepts a component which can be a string or
+  * a React Component and returns
+  * a "glamorousComponentFactory"
+  * @param {String|ReactComponent} comp the component to render
+  * @param {Object} options helpful info for the GlamorousComponents
+  * @return {Function} the glamorousComponentFactory
+  */
+  function glamorous(comp, {rootEl, displayName, forwardProps = []} = {}) {
     return glamorousComponentFactory
 
     /**
@@ -97,11 +101,15 @@ export default function createGlamorous(splitProps) {
             theme,
             this.context,
           )
+          const debugClassName = glamorous.config.useDisplayNameInClassName ?
+            cleanClassname(GlamorousComponent.displayName) :
+            ''
+          const className = `${fullClassName} ${debugClassName}`.trim()
 
           return React.createElement(GlamorousComponent.comp, {
             ref: props.innerRef,
             ...toForward,
-            className: fullClassName,
+            className,
           })
         }
       }
@@ -189,4 +197,8 @@ export default function createGlamorous(splitProps) {
       comp :
       comp.displayName || comp.name || 'unknown'
   }
+}
+
+function cleanClassname(className) {
+  return className.replace(/ /g, '-').replace(/[^A-Za-z0-9\-_]/g, '_')
 }
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,8 @@ const glamorous = createGlamorous(splitProps)
 Object.assign(
   glamorous,
   domElements.reduce((getters, tag) => {
+    // TODO: next breaking change, let's make
+    // the `displayName` be: `glamorous.${tag}`
     getters[tag] = glamorous(tag)
     return getters
   }, {}),
--- a/README.md
+++ b/README.md
@@ -499,10 +499,15 @@ const myGlamorousComponentFactory = glamorous(
 )
 ```
 
+> Note: the `displayName` can also included in the className that your
+> components are given which makes the development experience a bit nicer.
+> To enable this see the section about `config`.
+> This will likely be enabled by default in the next major change.
+
 And now all components created by the `myGlamorousComponentFactory` will have
 the `displayName` of `MyGlamorousComponent`.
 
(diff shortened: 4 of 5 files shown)


src/__tests__/__snapshots__/index.js.snap
```diff
@@ -7,11 +7,17 @@ exports[`allows you to pass custom props that are allowed 1`] = `
...
 <div
-  class="css-1r7lsit"
+  class="css-1r7lsit glamorous(mockConstructor)"
   id="hello"
 />
 `;
 
+exports[`allows you to specify a displayName 1`] = `
+<div
+  class="css-nil HiThere"
```
kentcdodds: Here's the real user-experience win right here. If the component has a `displayName` then we get that right here. And if folks are using the babel plugin, then that should work great.

src/__tests__/__snapshots__/index.js.snap
```diff
@@ -116,7 +122,7 @@ exports[`merges composed component styles for reasonable overrides 1`] = `
 }
 
 <div
-  class="other classes are not removed css-1k8eh41"
+  class="other classes are not removed css-1k8eh41 glamorous(glamorous(glamorous(div)))"
```
kentcdodds: Sometimes these can be rather funny 😅

src/create-glamorous.js
```diff
@@ -101,7 +101,7 @@ export default function createGlamorous(splitProps) {
           return React.createElement(GlamorousComponent.comp, {
             ref: props.innerRef,
             ...toForward,
-            className: fullClassName,
+            className: `${fullClassName} ${GlamorousComponent.displayName}`,
```
kentcdodds: This is the actual change. Pretty simple!

kentcdodds:
One other thing to consider. Does anyone have any problem with this happening in prod? I don't personally, but I could see why some folks might... 🤔

kentcdodds:
Also, this will definitely bust people's snapshots. Probably ok, but might ruffle some feathers.

src/create-glamorous.js
liadbiz: what is difference between `glamorous.div` and `glamorous(div)`?

src/__tests__/__snapshots__/index.js.snap
```diff
@@ -202,7 +208,7 @@ exports[`with theme prop of padding 10px 1`] = `
   }
 >
   <div
-    className="css-1p841yy"
+    className="css-1p841yy glamorous(div)"
```
liadbiz: what is difference between `glamorous(div)` and `glamorous.div`?

kentcdodds: this is just the `displayName`, so it's just a string that's automatically created for the displayName of components made with `glamorous.div()`. It's just a way to identify components that were made with that API.

kentcdodds:
Do we need to be worried about cleaning the `displayName` value so it works in a className? I don't know whether escaping is needed/necessary. Does react clean it up for us? Are there characters that aren't allowed in class names?

src/index.js
```diff
@@ -18,7 +18,7 @@ const glamorous = createGlamorous(splitProps)
 Object.assign(
   glamorous,
   domElements.reduce((getters, tag) => {
-    getters[tag] = glamorous(tag)
+    getters[tag] = glamorous(tag, {displayName: `glamorous.${tag}`})
```
kentcdodds: Note this change is inspired by @liadbiz who found `glamorous(div)` confusing. I think this is less confusing because it looks like how you would code it: `glamorous.div` rather than `glamorous(div)`.

src/__tests__/__snapshots__/index.js.snap
```diff
@@ -116,7 +122,7 @@ exports[`merges composed component styles for reasonable overrides 1`] = `
 }
 
 <div
-  class="other classes are not removed css-1k8eh41"
+  class="other classes are not removed css-1k8eh41 glamorous_glamorous_glamorous_div__"
```
kentcdodds: When you wrap a glamorous component multiple times like this you're `displayName` will be: `glamorous(glamorous(glamorous.div))` which isn't super easy to read. Unfortunately, because you cannot have `(` and `)` in class names, so this is even harder to read. However, many folks will hopefully be using the babel plugin so it should be more helpful :)

aaronjensen:
I don't think this is actually the same thing, is it? In styled-components the class names they generate have the component. So, instead of `css-1k8eh41` it would be `glamorous-div-1k8eh41` and then w/ the babel plugin, I would hope `MyComponent-1k8eh41`. 

Or maybe I'm misunderstanding something...?

kentcdodds:
Correct, this is not the same thing. To accomplish the same thing will require https://github.com/threepointone/glamor/issues/238 which will eventually be implemented. This is just to get us somewhere and relieve some pressure on @threepointone who is really busy with life right now :)

src/__tests__/__snapshots__/index.js.snap
```diff
@@ -7,11 +7,17 @@ exports[`allows you to pass custom props that are allowed 1`] = `
...
 <div
-  class="css-1r7lsit"
+  class="css-1r7lsit glamorous_mockConstructor_"
   id="hello"
 />
 `;
 
+exports[`allows you to specify a displayName 1`] = `
+<div
+  class="css-nil Hi-There"
```
kentcdodds: Most folks should see this if they're using the babel plugin :+1:

src/index.js
naknode: I agree. This seems less confusing to me as well as it appears "synthetic"

src/__tests__/__snapshots__/index.js.snap
marzelin: Instead of  `glamorous_glamorous_glamorous_div__` we could display `3*glamorous_div__`

marzelin: or even better `3*Glamorous-Div`

marzelin: I wonder why we cannot have parentheses in class names?

kentcdodds: Yeah, I'm not sure why the rules are the way they are, but yeah, you can only have certain characters in the class name of an element.

I think we can discuss things more in another issue if we want to change the default display name of glamorous components.

kentcdodds:
I've been thinking about this more and I'd rather not break everyone's snapshots. So I'm going to make this opt-in via some global config. I normally don't like global config, but in this case I think it's acceptable because this is intended only for debugging, so that should be fine.

I'm planning on something like:

```javascript
glamorous.config.useDisplayNameInClassName = true
```

In a future breaking version, we could change it to be opt-out.

src/__tests__/__snapshots__/theme-provider.js.snap
```diff
@@ -127,11 +127,11 @@ exports[`with theme prop of padding 20px 1`] = `
       }
     }
   >
-    <glamorous(div)>
+    <glamorous.div>
       <div
         className="css-1e41o48"
       />
-    </glamorous(div)>
+    </glamorous.div>
```
marzelin: The new name is more intuitive, but is it worth major version?
This change makes no difference to people who use awesome displayname plugin which provides much better debugging experience.

kentcdodds: Yeah, maybe I'll hold off on this change until we're going to make a major version bump anyway :+1:

PR #143, opened 2017-05-27
fix(exports): replace `const` with `var`

ethangodt:
**What**:
Replaces `'const'` with `'var'`.

**Why**:
To not break apps loading in environments without const support (since the code doesn't get transpiled).

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A — they pass
- [ ] Code complete N/A


kentcdodds:
Thanks!

PR #147, opened 2017-05-27
feat(exports): export internal getGlamorClassName function

knpwrs:
**What**: Exporting the internal `getGlamorClassName` function.

**Why**: I'm working on testing glamorous components with [WebdriverIO](http://webdriver.io/), it would be very helpful to me to have generated classnames.

**How**: See **What**.

**Checklist**:

- [ ] Documentation
- [ ] Tests N/A
- [x] Code complete
- [ ] Added myself to contributors table (TBD)
- [x] Followed the commit message forma

**Other**:

I'm open to other suggestions on how to do what I want here. This also unfortunately exports an internal API which can make future breaking changes more difficult if you're shy about releasing major versions. Maybe we should make the `getGlamorClassName` function accept a single object argument which is destructured before merging this PR?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@ import withTheme from './with-theme'
 import ThemeProvider from './theme-provider'
 import createGlamorous from './create-glamorous'
 import splitProps from './split-props'
+import getGlamorClassName from './get-glamor-classname'
 
 const glamorous = createGlamorous(splitProps)
 
@@ -58,4 +59,4 @@ function capitalize(s) {
 glamorous.default = glamorous
 
 export default glamorous
-export {ThemeProvider, withTheme}
+export {ThemeProvider, withTheme, getGlamorClassName}


kentcdodds:
The recommendation for testing is to use data-test attributes instead of class names. When I get to a keyboard I can show you what I mean with some examples. I don't want to expose the function though. Perhaps we can turn this into a pull request to document the recommended method for testing.

knpwrs:
@kentcdodds I believe I understand what you're saying. Something like:

```jsx
// Component
const MyFancyButton = glamorous.button({
  border: '1px solid black', // so fancy
});

// Rendering
<MyFancyButton data-test="MySpecificFancyButton" />

// WebdriverIO
browser.element('[data-test="MySpecificFancyButton"]').
```

That makes perfect sense to me, and it makes granularly finding specific components easier. As for your suggestion, I'd be open to writing some documentation. Are there any other examples or edge cases to consider?

kentcdodds:
That's exactly it! Yes, if you could change this PR (or close it and make a new one) that'd be awesome. Probably something for the examples. See the contributing guidelines 😀 thank you!

knpwrs:
I'll open a new PR a bit later. Thanks!

PR #148, opened 2017-05-27
Support dynamic style function which returns a static custom class name

zillding:
**What**:
Add support to allow component factory take in functions which return string class names.

**Why**:
This makes it easier to add some custom string class names to components based on props.

**How**:
Check the type of returned result from style functions to determine the next step of processing.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


zillding:
@kentcdodds I am still going through the code base to see where I can add test. 😊  Any suggestion/help would be welcome. 🙇

zillding:
😊  Thanks for the review. Glad I can help~

Issue #133, opened 2017-05-25
Support custom props


kentcdodds:
So styled-components v2 was released with a cool new [`attrs` API](https://www.styled-components.com/docs/basics#attaching-additional-props).

With glamorous, in our examples, we say you can do:

```javascript
const Password = glamorous.input({})
Password.defaultProps = { type: 'password' }
```

That works great unless you want the value to be dynamic and derived from props. The example they have is a little unnecessary, but I could see a situation where you'd want to have the props you pass to a wrapped component be mapped to some other value. So I wonder whether folks think this would be useful or just make things more complex.

### Option 1: Leave things as they are:

```javascript
const ValidatedInput = glamorous(({validationType, ...rest}) => {
  let extraProps = {type: validationType}
  if (validationType === 'credit-card') {
    extraProps = {pattern: creditCardPattern}
  }
  return <input {...extraProps} {...rest} />
}, {rootEl: 'input'})({
  /* styles */
})
```

### Option 2: Add an API:

```javascript
const ValidatedInput = glamorous.input({/* styles */}).props(({validationType}) => {
  if (validationType === 'credit-card') {
    return {pattern: creditCardPattern}
  }
  // validationType could be `email`, `url`, etc.
  return {type: validationType}
})
```

I value small APIs, but I also value ergonomics. I would love to hear people's perspectives on this. I'll open this up for a vote after we hear what folks have to say about this.

liadbiz:
IMO option2 is better because it's more consistent to style the component.
Only write css in api like glamorous.input(), and do not add custom property stuff to that.
if user want to add custom properties, just use another api like `.props` or `.attrs`

jfrolich:
or maybe:
```es6
import glamorous, { withProps } from 'glamorous'
const ValidatedInput = withProps(() => ())(glamorous.input({}))
```
An API that specifies the prop manipulation first before the glamorous component is a bit more intuitive for me because the prop manipulation happens first.

kentcdodds:
I like that. How about:

```javascript
import glamorous from 'glamorous'
const ValidatedInput = glamorous.input.withProps(props => ({}))(styles)
const ValidatedInput2 = glamorous(Input).withProps(propsFun)(styles)
```

I think that should be fairly straightforward :+1:

kentcdodds:
Let's do that. Who wants to work on this!? :smile:

marzelin:
This feature is super useful but is this a concern that a styling library should tackle? :thinking: 
[...]

Issue #145, opened 2017-05-27
withStyle helper method


marzelin:
Let's say you want a red button. What would be the most straight forward way you can think of?
What about:

```javascript
const Tomato = Button.withStyle({backgroundColor: 'tomato'})
```
It feels so natural, reads like English and does exactly what it says.

OK, you've got a boring tomato, let's make it more interesting:
```javascript
const HoverableTomato = Tomato.withStyle({'&:hover': {color: 'goldenrod'}})
```
Exactly the same method as before. Finally one simple and obvious way of doing things.

You think it's too good to be true? It's here! Check it out on [codesandbox.io](https://codesandbox.io/s/ERLK5X2nN)

Styling in this way feels so good. :)

ajwhite:
You can use the `glamorous` "factory factory" function 

```js
const Tomato = glamorous(Button)({backgroundColor: 'tomato'})
```
```js
const HoverableTomato = glamorous(Tomato)({'&:hover': {color: 'goldenrod'}})
```

And continue the HOC-inception 😃 

I do like the idea of `withStyle` though as a convenient way to accomplish the same thing

marzelin:
It's all about user convenience.
Why somebody would want to use this library? To improve their styling workflow.
What is a user is welcomed with? 17-page-long documentation.
I can imagine somebody looking at the docs and running away to their crappy but familiar workflow. "I just wanted add some styles, not get phd in computer science. Thankyouverymuch"

Let's get out of user's way as much as possible. Let them do the thing their want in as much straight-forward way there is.
Do you want an html element with some styles? Here you go: `HtmlElement.withStyle(someCssRules)`
 There's no mental translation needed. What you want is what you write. And inheriting from custom components is done in the same way - one thing less to learn and remember which converts to more time spent on actual work.

marzelin:
## Reasoning behind this change
### Interoperability
When you use current api, you have to keep in mind whether you're working with a `componentFactory` or `Component`. If you try to use `componentFactory` in jsx or try to style `Component` your application blows up.

With new api, when you import a built-in component
```javascript
import { Div } from 'glamorous'
```

You can use it directly in jsx
```javascript
<Div>Hello</Div>
```
Style it in jsx
```javascript
<Div css={{color: 'white'}}>Hello</Div>
```

Style it in js to benefit from reusability and inheritance
```javascript
const WhiteDiv = Div.withStyle({color: 'white'})
```
[...]

PR #149, opened 2017-05-27
feat(glamorous): withProps factory

ajwhite:
**What**: Adds `withProps` factory as discussed in #133

**Why**: A handy API for composing more than styles on a component 😃 

**How**: `glamorousComponentFactory` comes with a `withProps` property on it. This allows you to recompose the factory with properties

```js
const GlamorousComponent = glamorous(/* Comp */).withProps({ /* props */ })({ /* styles */ })
```

Note: these properties take a lower priority than properties defined on the `GlamorousComponent` by the user. Just like the styles.

## Feedback Requested

Before finalizing documentation, I'd like to get a sense of the approach is what we'd like!

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] ~~Added myself to contributors table~~ 

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,39 @@ const MyStyledDiv = glamorous.div(
 <MyStyledDiv /> // styles applied: {padding-top: 1, padding-right: 2, padding-bottom: 3, padding-left: 4} and anything coming from `extra-thing`.
 ```
 
+##### withProps
+
+The `glamorousComponentFactory` has a `withProps` property, which is another
+factory function accepting any number of property object arguments to be applied
+to the `GlamorousComponent`. Just like styles, these can either be property objects
+or functions wich are invoked with `props` on every render and return property objects.
+The result is a `GlamorousComponent` with the specified props:
+
+```jsx
+const StyledTextInput = glamorous.input.withProps({type: 'text'})({
+  padding: 10
+})
+
+const StyledValidatedTextInput = glamorous(MyStyledInput).withProps(
+  props => {
+    if (props.creditCard) {
+      return {
+        pattern: /([0-9]{4}){4}/
+      }
+    }
+    return null
+  }
+)
+
+<StyledTextInput />
+// renders: <input type="text" />
+// styles applied: {padding: 10}
+
+<StyledValidatedTextInput creditCard />
+// renders: <input type="text" pattern="/([0-9]{4}){4}/" />
+// styles applied: {padding: 10}
+```
+
 #### GlamorousComponent
 
 The `GlamorousComponent` is what is returned from the `glamorousComponentFactory`.
--- a/src/__tests__/index.js
+++ b/src/__tests__/index.js
@@ -433,3 +433,39 @@ test('should accept user defined contextTypes', () => {
   const props = {}
   expect(dynamicStyles).toHaveBeenCalledWith(props, theme, context)
 })
+
+test('withProps composes the component with the provided props', () => {
+  const Input = glamorous.input.withProps({type: 'text'})({color: 'red'})
+  expect(render(<Input />)).toMatchSnapshotWithGlamor()
+})
+
+test('withProps should accept a function producing props', () => {
+  const propsFn = ({validationType}) => {
+    const extraProps = {type: validationType}
+
+    if (validationType === 'credit-card') {
+      extraProps.pattern = /[0-9]/
+    }
+
+    return extraProps
+  }
+  const Input = glamorous.input.withProps(propsFn)({color: 'red'})
+  expect(
+    render(<Input validationType="credit-card" />),
+  ).toMatchSnapshotWithGlamor()
+})
+
+test('withProps is variadic', () => {
+  const propsFn = jest.fn()
+  const propsFn2 = jest.fn()
+
+  const Input = glamorous.input.withProps(propsFn, propsFn2)()
+
+  const wrapper = mount(<Input type="text" pattern="foo" />)
+  expect(propsFn).toHaveBeenCalledWith(wrapper.props())
+})
+
+test('withProps should have a lower priority than component props', () => {
+  const Input = glamorous.input.withProps({type: 'text'})({color: 'red'})
+  expect(render(<Input type="number" />)).toMatchSnapshotWithGlamor()
+})
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -27,6 +27,7 @@ function createGlamorous(splitProps) {
   * @return {Function} the glamorousComponentFactory
   */
   function glamorous(comp, {rootEl, displayName, forwardProps = []} = {}) {
+    glamorousComponentFactory.withProps = withProps
     return glamorousComponentFactory
 
     /**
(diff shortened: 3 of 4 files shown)


src/__tests__/index.js
```diff
@@ -433,3 +433,29 @@ test('should accept user defined contextTypes', () => {
   const props = {}
   expect(dynamicStyles).toHaveBeenCalledWith(props, theme, context)
 })
+
+it('should compose a component withProps', () => {
```
kentcdodds: I prefer `test('composes a component...`

ajwhite:
Would love some 👀 from @jfrolich too!

src/create-glamorous.js
```diff
@@ -161,6 +159,45 @@ function createGlamorous(splitProps) {
...
+     * This returns a glamorousComponentFactory that will be composed with
+     * the provided properties.
+     *
+     * @param {...Object|Function} outerProps the properties to provide to the
+     *  GlamorousComponent.
+     *  If any of these are functions, they will be invoked with the component
+     *  props and the return value is used.
+     * @return {Function} the glamorousComponentFactory function.
+     */
+    glamorousComponentFactory.withProps = (...outerProps) => {
```
kentcdodds: I prefer function declarations for all functions. Then above we can do:

```
glamorousComponentFactory.withProps = withProps
return glamorousComponentFactory
```

src/create-glamorous.js
```diff
@@ -161,6 +159,45 @@ function createGlamorous(splitProps) {
...
+      // This is a higher order function of `glamorousComponentFactory`
+      return (...styles) => {
```
kentcdodds: Let's use a function declaration here too. That way this function can have a name which will be used in stack traces 😀

src/create-glamorous.js
ajwhite: Great point!

src/create-glamorous.js
```diff
@@ -161,6 +162,49 @@ function createGlamorous(splitProps) {
...
+        // this is a higher order component of the GlamorousComponent that will
+        // evaluate the properties from withProps.
+        return function GlamorousComponentWithProps(innerProps) {
+          // take the properties provided to the GlamorousComponent and evaluate
+          // the properties from `withProps`
+          const evalutatedOuterProps = outerProps.reduce((props, prop) => {
+            const evaluated = typeof prop === 'function' ?
+              prop(innerProps) :
+              prop
+            return {...props, ...evaluated}
```
ajwhite: I forget -- if we attempt to spread `evaluated`, and it evaluated as `null` or `undefined` (or any non object, really) -- that would raise an exception wouldn't it?

We should probably ensure this is an object, and perhaps test that it's a non-empty object?

kentcdodds: The values of evaluated and props should always be objects. But we should do a falsy check here.

ajwhite: Agreed, but just like the dynamic styles, it sounds good to safeguard scenarios where the user may have a conditional return value, which could result in `null`

```js
glamorous.input(props => {
  if (props.foo) {
    return {color: 'blue'}
  }
  return null
})
```
```js
glamorous.input.withProps(props => {
  if (props.foo) {
    return {type: 'number'}
  }
  return null
})
```

ajwhite: Ah, by "should always be an object", I realize you mean that it shouldn't be an array, string, etc -- you're not saying that `null` isn't a valid return value

README.md
```diff
@@ -299,6 +299,39 @@ const MyStyledDiv = glamorous.div(
 <MyStyledDiv /> // styles applied: {padding-top: 1, padding-right: 2, padding-bottom: 3, padding-left: 4} and anything coming from `extra-thing`.
 ```
 
+##### withProps
```
kentcdodds: I just had a thought. Do you think that people may get confused and try to do:

```javascript
const MyStyledInputWithProps = glamorous.input({
  padding: 10
}).withProps({type: 'text'})

// or maybe:

const MyStyledInput = glamorous.input({
  padding: 10
})

const MyStyledInputWithProps = MyStyledInput.withProps({type: 'text'})
```

In interested of that, perhaps we should support both?

README.md
```diff
@@ -299,6 +299,39 @@ const MyStyledDiv = glamorous.div(
...
+    return null
+  }
+)
+
+<MyStyledInput />
+// renders: <Input type="text" />
+// styles applied: {padding: 10}
+
+<MyStyledValidatedInput creditCard />
+// renders: <Input type="credit-card" pattern="/([0-9]{4}){4}/" />
```
kentcdodds: This should render `<input` rather than `<Input`, also I'm pretty sure that `credit-card` is not a valid `type`. Which is why the pattern is good :)

ajwhite:
Docs are added, let me know if there's anything I can improve on those

README.md
ajwhite: I think supporting both would be convenient - would your concern in https://github.com/paypal/glamorous/issues/145#issuecomment-304471670 apply here though?

README.md
ajwhite: Ha good catch

README.md
kentcdodds: Yeah, I've been thinking about it more and I think that I'm ok with doing this after all...

ajwhite: It follows similar patterns as in librarys like ImmutableJS, where you use the core API to _create the subject_, and you can build changes from working against the _subject's API_ itself, and not necessarily on the core's factory API

```js
var map = Immutable.Map({foo: 'bar'}) // factory API
var map2 = map.set('bar', 'foo') // subject API
```

So, similarly:
```js
var Component = glamorous.div()
var Component2 = Component.withProps({foo: 'bar'})
```

ajwhite: I actually like this, as it opens the door for us to work with both:

```js
var Component = glamorous.div({backgroundColor: 'blue'})
var Component2 = Component.withProps({foo: 'bar'})
var Component3 = Component2.withStyle({color: 'red'})
```

Giving a consistent behavior to `with*` APIs

ajwhite: I still want to think about this before committing, maybe bring in a few people to vote. 

I like both, but they introduce different conventions. Should we be adding the API to _components_, or keep it isolated to the glamorous factory?

There's traits about both approaches that I like. Might pause on this and write up my thoughts comparing both approaches

kentcdodds:
I think we're going to opt for an integration example with `recompose` rather than implementing it as a built-in. Unless I'm missing something, that should be enough right?

ajwhite:
`recompose` looks _fantastic_. I don't see a huge benefit for incorporating this after seeing their library, which provides _all_ the utility you might need to compose a HOC. Slippery slope - right? If we do `withProps`, where does it end? `withState` becomes asked for next?

I agree that the right move here is to promote the usage of `recompose` for situations like this,  an example of [`recompose.mapProps`](https://github.com/acdlite/recompose/blob/master/docs/API.md#withprops) would be terrific.

I'm going to close this 👍

kentcdodds:
Ok, with #175 done, I think we're ready to give this another go. Based on discussion in #133, I'd like to support this API:

```js
import glamorous, { withProps } from 'glamorous'

const withPropsObject = {propA: 'a', propB: 'b'}

// withProps functions accept the same arguments that glamorous style functions accept.
const withPropsFunction = (props, theme, context) => ({propG: 'g', propH: 'h'})

// we'll call these withPropsThing, meaning when you see withPropsThing it could be either a function or object

const Div1 = glamorous.div({}).withProps(withPropsThing)
<Div1 /> // renders div with the props, filters out invalid `div` props like we do currently.
```

What do we need to do to make this happen? How can I help you @ajwhite?

ajwhite:
The most actionable item here is to expose `theme` and `context` to the `withProps` component wrapper. Happy to accept contributions to the existing work, otherwise I can likely turn something around within the next few days

kentcdodds:
How can we help get this finished? I'd love to release this with the next major version

kentcdodds:
I'm going to go ahead and finish this :+1:

kentcdodds:
This is superseded by #255

ajwhite:
Thanks for bringing this across the finish line, been a busy week and was on vaca!

kentcdodds:
Sure thing! Thanks for all the work you've done!

PR #154, opened 2017-05-29
fix(typings): added missing declaration for withTheme()

debradley:
**What**: Added function typing for withTheme
**Why**: It didn't exist
**How**: Added declaration to typings/glamorous.d.ts

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete

As a TypeScript noob I have no idea if my function declaration is accurate or optimal. Feedback needed.


kentcdodds:
Thanks! I don't know TypeScript, so I'm not much use when it comes to reviewing PRs. But a few other folks will give this a look for you. Thanks again for the PR :)

typings/glamorous.d.ts
```diff
@@ -40,6 +40,12 @@ interface ThemeProps {
 
 export class ThemeProvider extends React.Component<ThemeProps, any> { }
 
+export declare function withTheme
+  <P>(
+    component:
+      | React.ComponentClass<P>
```
sammkj: This would fail when user tries to use `theme` prop. To avoid that, it could be written as

```
export type Component<T> = React.ComponentClass<T> | React.StatelessComponent<T>;

// `ThemeProps` comes from https://github.com/paypal/glamorous/blob/master/typings/glamorous.d.ts#L37-L39.
export declare function withTheme<P>(component: Component<P & ThemeProps>): GlamorousComponent<P>;

```

sammkj:
@debradley thanks for adding new definitions! Everything looks good, just one thing missing. Let me know if you have any questions!

typings/glamorous.d.ts
debradley: Your definition of Component<T> looks like it could also be used in the current `GlamorousInterface`. Does this refactoring look correct? (I'm brand new to TypeScript, I honestly don't understand the generics or many other things yet).

```js
export type Component<T> = React.ComponentClass<T> | React.StatelessComponent<T>;

export interface GlamorousInterface extends HTMLGlamorousInterface, SVGGlamorousInterface {
  <P>(
    component:Component<P>,
    options?: GlamorousOptions
  ): StyledFunction<P, CSSProperties | React.SVGAttributes<any>>

export declare function withTheme
  <P>(component: Component<P & ThemeProps>): GlamorousComponent<P>;
```

ErikCupal:
Great job :)) not home currentcly, I'll review the PR later today ;)

typings/glamorous.d.ts
sammkj: @debradley Looks good! Can you also add a test case in https://github.com/paypal/glamorous/blob/master/test/glamorous.test.tsx ?

debradley: The method of testing you're using there is completely opaque to me and I'm not sure what a valid test is. And again, I'm completely TypeScript-illiterate. So I've taken a guess but TS isn't happy with it and I can't stumble on something that works... Here's what I tried:

```js
import glamorous, { withTheme, ThemeProvider } from "../";
// ...
const ThemedClass = withTheme(ClassToWrap)({})
```

But I get: "Value of type 'ComponentClass<object & ExtraGlamorousProps>' is not callable. Did you mean to include 'new'?"

Suggestions?

sammkj: Ahh I see. you have to remove the function call. Instead of `withTheme(ClassToWrap)({})`, it should be `withTheme(ClassToWrap)`.

I gotta admit this is not the best way to write test cases to validate the Typescript definitions and it's very messy. If you stumble upon anything better, please do let us know!

debradley: I've committed that fix and a test for your review.

kentcdodds:
Awesome! Great work!

PR #155, opened 2017-05-30
docs(README): fix TOC link issue

kentcdodds:
**What**: Changes `# glamorous` to `<h1>glamorous</h1>`

**Why**: Closes #150 

**How**: update to README.md

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete 
- [ ] Added myself to contributors table N/A
- [x] Followed the commit message format

PR #157, opened 2017-05-30
Update README - fix anchor to Glamorous heading

kylegach:
**What**: The ‘Glamorous’ anchor in the table of contents points to the top-level heading instead of the expected heading lower down.

**Why**: GitHub apparently numbers identical headings, so this change makes the ‘Glamorous’ anchor link in the TOC work as expected.

**How**: Edited the markdown; previewed the fix on GitHub.

**Checklist**:

- [x] Documentation
- [x] Tests N/A
- [x] Code complete N/A
- [ ] Followed the commit message format  (_Sorry, I'm not sure how to do that within the GitHub UI._)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ React component styling solved with an elegant ([inspired](#inspiration)) API,
   - [Typescript](#typescript)
 - [Video Intro :tv:](#video-intro-tv)
 - [Terms and concepts](#terms-and-concepts)
-  - [glamorous](#glamorous)
+  - [glamorous](#glamorous-1)
   - [glamorous API](#glamorous-api)
   - [Theming](#theming)
   - [Config](#config)


kentcdodds:
Ah! Nice! Thank you for this. We'd decided this was a wont fix in #155. The problem is that the table of contents is generated. So this change will get overwritten. So this sounds like a bug in our generator [`doctoc`](https://github.com/thlorenz/doctoc). Perhaps we could make a pull request over there to fix it?

Unfortunately, we're not going to be able to do anything reasonable in this repo to actually fix this long-term, so I'll go ahead and close this. Thanks though!

kylegach:
Gotcha. I saw that other PR, but didn’t realize the TOC was generated.

kentcdodds:
Thanks anyway @kylegach!

Issue #135, opened 2017-05-25
Improve changing the underlying component that's rendered


kentcdodds:
In the [v2 styled-components announcement blogpost](https://medium.com/styled-components/announcing-v2-f01ef3766ac2), they introduced a new API called the `.extend` helper. We actually do not need this because our `glamorous(Component)` already behaves in this way. So that's a relief (however we may be able to improve docs so this is more clear 🤔).

But they also introduced a `.withComponent` API which I think is kinda handy:

```javascript
const Heading = styled.h1`
  /* ... */
`
export const H1 = Heading
export const H2 = Heading.withComponent('h2')
export const H3 = Heading.withComponent('h3')
export const LinkHeading = Heading.withComponent(Link)
```

For us, to do this now is not really ergonomic:

```javascript
const Heading = glamorous.h1({})

const H2 = glamorous.h2(Heading.styles)
```

Should we add a `withComponent` API?

```javascript
const Heading = glamorous.h1({})
const H2 = Heading.withComponent('h2')
```

Please leave a :+1: (yes) or :-1: (no, don't add the new API)

kentcdodds:
Ok, so I created [this](https://codesandbox.io/s/X6zjo7o1V) as a proof of concept way to do this without building something into glamorous. However, it's a bit of a hack _and_ what's worse is it results in a full render of the original component which _could_ have unintended side effects.

So I think that we'll want this built-into glamorous 👍  Anyone wanna work on this? [Here's](https://github.com/paypal/glamorous/blob/master/src/create-glamorous.js#L109) where we render the component. We'll basically want to add to [this `Object.assign` call](https://github.com/paypal/glamorous/blob/ea52e0f72d45d8e0901ae608c2c1f68e8dfe3a2a/src/create-glamorous.js#L152-L161) and we'll want that to add a function called `withComponent` which accepts a component and assigns the value of `GlamorousComponent.comp` and even `GlamorousComponent.rootEl` I think. I'm thinking that the `withComponent` API should be similar to [the `glamorous` function API](https://github.com/paypal/glamorous/blob/ea52e0f72d45d8e0901ae608c2c1f68e8dfe3a2a/src/create-glamorous.js#L29): `withComponent(comp, {rootEl = comp.comp ? comp.comp : comp, displayName = `glamorous(${getDisplayName(comp)})`, forwardProps = []} = {})`. Then overriding those properties on `GlamorousComponents`.

[...]

PR #159, opened 2017-06-01
feat(glamorous): add withComponent api

ansumanshah:
**What**: Adds withComponent function as discussed in #135 

**Why**: for changing the underlying component that's rendered :)

**How**: 
```js
function withComponent(newComp, options) {
  return glamorous(newComp, {
    ...options,
    forwardProps: GlamorousComponent.forwardProps,
  })(GlamorousComponent.styles)
}
```

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] Followed the commit message format


src/create-glamorous.js
```diff
@@ -149,6 +149,15 @@ function createGlamorous(splitProps) {
         },
       })
 
+      function withComponent(newComp) {
+        return Object.assign(GlamorousComponent, {
```
kentcdodds: Ah yes! We don't want to be manipulating the existing component, we want to make a _new_ one that has the same properties! So instead this function should probably be:

```javascript
function withComponent(newComp, options) {
  return glamorous(newComp, {
    ...options,
    forwardProps: GlamorousComponent.forwardProps,
  })(GlamorousComponent.styles)
}
```

I _think_ that should do it....

ansumanshah: Thanks a lot.. I was just not attaching styles the right way :( :(

kwelch:
I really like this and saw no issues in the code, but I do have one question. 

Is it possible to add more styles to the resulting component?

The example in the docs show styling a `Button` and then using those styles for a `Link`, but by default an `a` tag is going to have `text-decoration` which is not a concern for a button. 

Not sure is this is meant to accomplish that since their are others means to do that such as 
```
const buttonStyles = { ... };
const Button = glamorous.button(buttonStyles);
const Link  = glamorous.a({ ...buttonStyles, textDecoration: 'none' });
```

I don't think this is a blocking thing, but just wanted to raise the question.

kentcdodds:
Let's actually add that. I think that'd be really useful. I'd like to be able to do:

```js
const buttonStyles = { ... };
const Button = glamorous.button(buttonStyles);
const Link  = Button.withComponent('a', { styles: textDecoration: 'none' });
```

Or what about:

```js
const buttonStyles = { ... };
const Button = glamorous.button(buttonStyles);
const Link  = Button.withComponent('a', {/* options must be present */}, ...styles);
```

I like the second one better because the API is familiar, but often you wont have any options so including an empty object could be annoying...

kwelch:
How do you think you would handle the dynamic styles? (styles as a function, not sure the name)

kentcdodds:
I made an update to the code example above. Wouldn't complicate the implementation any further

ansumanshah:
I would go with the first example and maybe modify the API a little bit to make way for styles object inside the options one
```js
const buttonStyles = { ... };
const Button = glamorous.button(buttonStyles);
const Link  = Button.withComponent('a', { styles: {textDecoration: 'none'}, /* options if any */ });
```

ajwhite:
I don't love the interfaces of the current proposals ☹️  I think that's partially why this PR is somewhat stalled - none of us are _in love_ with it and completely sold one way or the other.

I fully understand why they are the way they are -- we need to allow for `glamorous` options, and styles are variadic so they must be last in the argument list, which it makes it difficult to have a simple interface.

I'm going to drop a few alternatives here while we think about this, none of which are perfect either.

----

### `withComponent` as a component factory

```js
const Link = Button.withComponent('a', { /* options */ })({ /* styles */ })
```
#### Upsides
- Interface is the same as `glamorous(component, options)(...styles)`
- Reusability by creating a factory

#### Downsides
- Less ergonomic, as we're now curried and it makes it harder to compose a new component when you don't have additional options:
  ```js
  const Link = Button.withComponent('a')()
  ```

----

### `withComponentOptions` as an expanded interface

Support `withComponentOptions` additionally to `withComponent`

```js
// `withComponent` interface is strictly styles
const Link = Button.withComponent('a', ...styles)

// `withComponentConfiguration` interface is options + styles
const ConfiguredLink = Button.withComponentOptions('a', options, ...styles)
```

#### Upsides
- Cleaner interface - use only what you ask for

#### Downside
- Larger API surface area - using different functions to dictate the interface

-----

Out of these two options, I like the component factory approach more. 

Out of _all_ the proposals in this PR, I'm juggling between:
- `Button.withComponent('a', {/* options must be present */}, ...styles)`
- `Button.withComponent('a', options)(...styles)`

kentcdodds:
What about this:

```javascript
const Button = glamorous.button({})
const Anchor = Button.withComponent('a', options)
// this is probably as much as most people want. If they want to add styles:
const MoreStyledAnchor = glamorous(Anchor)(...styles)

// want to do that in a single line?
const Anchor = glamorous(Button.withComponent('a'))(...styles)
```

I think I like that the most. So we don't allow you to specify more styles in the `withComponent` API directly. I don't expect that will be a very common use case anyway. And if you want to, it's pretty easy to just wrap it in a `glamorous` call.

So why don't we consider which would be the most common use case?

1. `Button.withComponent('a', {/* options must be present */}, ...styles)`
2. `Button.withComponent('a', options)(...styles)`
3. `Button.withComponent('a', ...styles)` and `Button.withComponentOptions('a', options, ...styles)`
4. `Button.withComponent('a', options)` with docs indicating you can rewrap that in `glamorous` for additional styles.

I favor option # 4 most.

kentcdodds:
I'm going to push forward with option # 4. I'll push an update to this branch.

ajwhite:
Option 4 looks good!

Simple enough:
```js
const MyH1 = glamorous.h1({...})
const MyH2 = glamorous(
  MyH1.withComponent('h2')
)({...})
```

kentcdodds:
Yep! And I expect most of the time folks wont care to alter the CSS anyway. And even if they do, there are multiple ways to do it:

```javascript
const MyH1 = glamorous.h1({...})
const MyH2 = MyH1.withComponent('h2')
// this
const MyStyledH2 = glamorous(MyH2)({...})
// or
<MyH2 css={{...}} />
// or
<MyH2 className={glamor.css({...})} />
```

Pretty flexible :)

kentcdodds:
For some reason I couldn't edit this PR, so I opened [another one](https://github.com/paypal/glamorous/pull/168) that's ready for review and merge! No worries @ansumanshah, your commit is still in that one! You wont be lost in the history :)

PR #160, opened 2017-06-02
feat(typescript): Improved theme typing and tests

luke-john:
BREAKING CHANGE: The updated theme typing will complain if existing usage of theme is unsafe

**What**:
Updated the typings around themes to get better safety around usage

**Why**:
We're implementing theming in a component library and wanted safety.

**How**:
Updated the typescript definition files and tests.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/test/glamorous.test.tsx
+++ b/test/glamorous.test.tsx
@@ -12,34 +12,38 @@ const Static = glamorous.div({
 });
 
 // dynamic styles
-const Title = glamorous.h1(
+const Title = glamorous.h1<{ color: string }, {}>(
   {
     "fontSize": "10px",
     "zIndex": "auto",
   },
-  (props: { color: string; }) => ({
+  (props) => ({
     "color": props.color,
   }),
 );
 
+const UseTitle = () => (
+  <Title color="red" />
+)
+
 // theme styles
-const Divider = glamorous.span(
+const Divider = glamorous.span<{}, { main: { color: string; } }>(
   {
     "fontSize": "10px",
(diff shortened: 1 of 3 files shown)


ErikCupal:
Great! 😃

Since Typescript 2.3 it's possible to set default generics type https://github.com/Microsoft/TypeScript/pull/13487. It's useful here as it cuts off a little boilerplate. See:

```ts
// Without the default generics

export interface StyledFunction<Props, Properties> {
  <CustomProps, ThemeProps>(
    // ...
  ): GlamorousComponent<Props & CustomProps>
}

...

const Title = glamorous.h1<{ color: string }, {}>(
  {
    "fontSize": "10px",
    "zIndex": "auto",
  },
  (props) => ({
    "color": props.color,
  }),
);

// With the default generic

export interface StyledFunction<Props, Properties> {
  <CustomProps, ThemeProps = {}>(
    // ...
  ): GlamorousComponent<Props & CustomProps>
}

...

const Title = glamorous.h1<{ color: string }>(
  {
    "fontSize": "10px",
    "zIndex": "auto",
  },
  (props) => ({
    "color": props.color,
  }),
);
```

I'll commit the changes for this improvement

kentcdodds:
Thanks friends! So I'd like to avoid a major version bump if I can. At least, if we're doing that we should probably do other breaking changes while we're at it. So I'm going to merge this into `next` and we can talk about other things we can do.

PR #161, opened 2017-06-02
feat(typescript): Improved theme typing and tests (#160)

kentcdodds:
This is our BREAKING CHANGE branch. Once this is merged, a new major release will go out. What else do we want to group into this?

### DIFF
--- a/test/glamorous.test.tsx
+++ b/test/glamorous.test.tsx
@@ -12,34 +12,38 @@ const Static = glamorous.div({
 });
 
 // dynamic styles
-const Title = glamorous.h1(
+const Title = glamorous.h1<{ color: string }>(
   {
     "fontSize": "10px",
     "zIndex": "auto",
   },
-  (props: { color: string; }) => ({
+  (props) => ({
     "color": props.color,
   }),
 );
 
+const UseTitle = () => (
+  <Title color="red" />
+)
+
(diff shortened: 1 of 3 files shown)


luke-john:
The excellent addition to the PR from @ErikCupal actually meant that that change is not really Breaking, but more of a Patch.

Using the new language feature of default types on generics meant that we were able to get typesafety with themes without introducing boilerplate when using components that don't use the theme.

The only additional changes in the commit will only break existing code on an update if they have used glamorous incorrectly.

There may be some other similar changes that would be good to land, any that you're aware of @ErikCupal @sammkj?  I've got a long weekend and was planning on doing some work with glamorous on monday so would be happy to pick anything up then :).

kentcdodds:
Whelp, I'd rather not do a major release anyway, so if you're saying that it's unlikely to break anyone then let's just do a patch release. That ok?

ErikCupal:
Yep, the change is not breaking anymore. I agree with Kent, we should just do a patch release 🙂.

luke-john:
Sounds sensible to me, apologies for the extra work that commit led to for you.

kentcdodds:
I merged this manually to avoid an accidental major release (because GitHub's UI sometimes gets messed up when changing commit messages)

PR #171, opened 2017-06-09
feat(css): allow the `css` prop to accept fns and strings

### LINKED ISSUE #170: Allow `css` prop to accept a function
So we can use the theme in it

I realize that we can use the `withTheme` HoC, but I've got reusable functions that I would rather just stick right in the `css` prop position :)

I need this now so I'm going to work on it now :)

kentcdodds:
**What**: Closes #170

**Why**: enables code reuse and potentially "Ahead of Time compilation"

**How**:

- Reusing the same logic for `cssOverrides` as we use for
`styles` in `get-glamor-classname.js`. Tiny refactor there.
- Adding tests to cover all the cases for the API.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

I also updated deps because I think prettier was messed up.

### DIFF
--- a/src/umd-entry.js
+++ b/src/umd-entry.js
@@ -4,16 +4,13 @@ const glamorous = glamorousStar.default
 
 Object.assign(
   glamorous,
-  Object.keys(glamorousStar).reduce(
-    (e, prop) => {
-      if (prop !== 'default') {
-        // eslint-disable-next-line import/namespace
-        e[prop] = glamorousStar[prop]
-      }
-      return e
-    },
-    {},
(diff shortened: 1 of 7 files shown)


src/umd-entry.js
```diff
@@ -4,16 +4,13 @@ const glamorous = glamorousStar.default
 
 Object.assign(
   glamorous,
-  Object.keys(glamorousStar).reduce(
```
kentcdodds: prettier update. I didn't change anything here

PR #181, opened 2017-06-14
fix(css): improve built-in support with css prop as function

kentcdodds:
I'm working on fixing a bug I found while making [this](https://codesandbox.io/s/ERNVNoxEv) (notice that the `textAlign="center"` is not applied to the first user card because the `css` prop is set to a function.

As part of making that an easier fix, I'm refactoring `getGlamorClassName` and I'm not sure why a bunch of the tests are breaking. I've only dug a little ways into it, but now I've got daddy duty. If anyone wants to debug it and figure out what's up, that'd be awesome. At first glance, it appears that the class names are generated properly, but somehow they're not added to glamor's stylesheet which is why they're not appearing in the snapshot???? Weird stuff.

### DIFF
--- a/src/__tests__/__snapshots__/index.js.snap
+++ b/src/__tests__/__snapshots__/index.js.snap
@@ -80,23 +80,23 @@ exports[`can use pre-built glamorous components with css attributes 1`] = `
 `;
 
 exports[`can use pre-built glamorous components with css prop 1`] = `
-.css-1n96jtr,
-[data-css-1n96jtr] {
+.css-1d3tm0b,
+[data-css-1d3tm0b] {
   -webkit-flex-direction: column;
   -ms-flex-direction: column;
   -webkit-box-orient: vertical;
   -webkit-box-direction: normal;
-  background: blue;
   display: -webkit-box;
   display: -moz-box;
   display: -ms-flexbox;
   display: -webkit-flex;
   display: flex;
   flex-direction: column;
+  background: blue;
 }
 
 <a
-  class="css-1n96jtr is added to classes"
+  class="css-1d3tm0b is added to classes"
   href="/is-forwarded"
   id="is-forwarded"
 />
@@ -114,13 +114,15 @@ exports[`css prop with a className 1`] = `
 `;
 
 exports[`css prop with a function 1`] = `
-.css-1xnhj1e,
-[data-css-1xnhj1e] {
+.css-1ps4aij,
+[data-css-1ps4aij] {
+  font-size: 10px;
   padding: 20px;
+  margin: 30px;
 }
 
 <section
-  class="css-1xnhj1e"
+  class="css-1ps4aij"
 />
 `;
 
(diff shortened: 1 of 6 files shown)


src/create-glamorous.js
```diff
@@ -94,13 +94,13 @@ function createGlamorous(splitProps) {
             Object.freeze(this.state.theme)
 
           // create className to apply
-          const fullClassName = getGlamorClassName(
```
kentcdodds: Just turning this into an object because there are a lot of args (and I'm going to add yet another)

src/get-glamor-classname.js
```diff
@@ -25,29 +25,21 @@ function extractGlamorStyles(className = '') {
 
 export default getGlamorClassName
 
-function getGlamorClassName(styles, props, cssOverrides, theme, context) {
+function getGlamorClassName({
```
kentcdodds: to-object refactor

src/get-glamor-classname.js
```diff
@@ -25,29 +25,21 @@ function extractGlamorStyles(className = '') {
...
+  context,
+}) {
   const {mappedArgs, nonGlamorClassNames} = handleStyles(
-    styles,
+    [...styles, ...(props.className || '').split(' '), cssOverrides],
     props,
     theme,
     context,
   )
-  const {
```
kentcdodds: I realized that most of this stuff was just calling into `handleStyles` and I could just call that once with the joined style args and that should give us the same effect. By "should," I mean, it would if the tests were passing 🙃

src/__tests__/__snapshots__/index.js.snap
```diff
@@ -73,7 +73,7 @@ exports[`can use pre-built glamorous components with css attributes 1`] = `
 }
 
 <a
-  class="is added to classes css-1t62idy"
+  class="css-1t62idy is added to classes"
```
kentcdodds: The order of the classes is not part of the supported API, so I'm ok with this.

kentcdodds:
Cool, I've got the build fixed, now to fix the core issue...

kentcdodds:
Hey people watching the repo (anyone), could I get a few reviews?

src/get-glamor-classname.js
```diff
@@ -25,31 +25,32 @@ function extractGlamorStyles(className = '') {
...
+  theme,
+  context,
+}) {
+  const {
+    glamorStyles: parentGlamorStyles,
+    glamorlessClassName,
+  } = extractGlamorStyles(props.className)
   const {mappedArgs, nonGlamorClassNames} = handleStyles(
-    styles,
+    [...styles, parentGlamorStyles, cssOverrides, cssProp],
```
kwelch: Is this determining precedence? If so, this is the order you expect them in? Are they reduced? I would expect styles to have highest importance.

kwelch:
Looks good from what I can see, but honestly not familiar enough. Not seeing anything glaring.

src/get-glamor-classname.js
kentcdodds: Yes, and this preserves the original presidence. We have tests for it which you can check out. I'd link you to them, but I'm on mobile.

kwelch: No it is fine. From what I saw the tests worked as I would have expected.

kentcdodds:
Thanks for the review @kwelch!

PR #183, opened 2017-06-14
Support a component as selector API

### LINKED ISSUE #136: Add the ability to use a component as a selector in CSS
In [the styled-components v2 blogpost announcement](https://medium.com/styled-components/announcing-v2-f01ef3766ac2) it says:

> You can now reference another component in your CSS, for cases where you really need to make a contextual override.

I think this could definitely be handy! @threepointone said that he'd be willing to make a PR to make this happen, because he's just really nice :)

kentcdodds:
**What**: This adds support for components as selectors

**Why**: To cover an edge case more nicely. Closes #136

**How**: use `toString` and `GlamorousComponent.toStringClassName`

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] ~Added myself to contributors table~ 

Still trying to figure out some stuff. Gotta do something like [this](https://twitter.com/mxstbr/status/875072178198388736). Help appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1003,6 +1003,48 @@ const MyResponsiveDiv = glamorous.div({
 ```
 </details>
 
+In addition to all this, you can reference a `glamorous` component inside the
+styles of other `glamorous` components (this works because they each implement
+the `toString` method):
+
+<details>
+<summary>Contextual Overrides</summary>
+
+This API is inspired by `styled-components`.
+
+```javascript
+const Link = glamorous.a({
+  color: 'inherit',
+  textDecoration: 'none',
+  '&:hover, &:active': {
+    textDecoration: 'underline',
+  }
+})
+
+const P = glamorous.p({
+  [Link]: {
+    textDecoration: 'underline',
+  }
+})
+
+render(
+  <P>
+    Hello!
+    <Link href="https://github.com/paypal/glamorous">
+      glamorous on GitHub
+    </Link>
+  </P>
+)
+// this will render glamorous on GitHub textDecoration as underline.
+
+```
+
+> Tip: in unit tests using snapshots where you don't want the glamorous IDs to
+> change every time you add a glamorous component between tests, you can reset
+> the `glamorous.config.idCounter = 1`.
+
+</details>
+
 ## Related projects
 
 - [babel-plugin-glamorous-displayname][babel-displayname]: Automatically adds a `displayName` to your glamorous components for a better debugging experience.
--- a/package.json
+++ b/package.json
@@ -96,12 +96,13 @@
     "rules": {
       "import/prefer-default-export": 0,
       "react/no-unused-prop-types": 0,
-      "valid-jsdoc": 0
+      "valid-jsdoc": 0,
+      "max-len": 0
     }
   },
   "lint-staged": {
     "*.js": [
-      "prettier-eslint --write",
+      "prettier-eslint --write --print-width=80",
       "git add"
     ],
     "README.md": [
--- a/src/__tests__/__snapshots__/index.to-string-class-name.js.snap
+++ b/src/__tests__/__snapshots__/index.to-string-class-name.js.snap
@@ -1,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`allows you to use a component as a child selector 1`] = `
+.css-3v2xme .g-1-1,
+[data-css-3v2xme] .g-1-1 {
+  color: blue;
+}
+
+.css-1ezp9xe,
+[data-css-1ezp9xe] {
+  color: red;
+}
+
+<div
+  class="css-3v2xme"
+>
+  <div
+    class="css-1ezp9xe g-1-1"
+  >
+    Hello
+  </div>
+</div>
+`;
+
+exports[`allows you to use a component as any kind of selector 1`] = `
+.css-icjsl7,
+[data-css-icjsl7] {
+  color: blue;
+}
+
+.g-1-1 .css-1ha6yxt,
+.g-1-1 [data-css-1ha6yxt] {
+  color: green;
+}
+
+<div
+  class="css-icjsl7 g-1-1"
+>
+  <div
+    class="css-1ha6yxt"
+  >
+    Hello
+  </div>
+</div>
+`;
--- a/src/__tests__/index.to-string-class-name.js
+++ b/src/__tests__/index.to-string-class-name.js
@@ -1,0 +1,47 @@
+/* eslint func-style:0, react/prop-types:0 */
+import React from 'react'
+import {render} from 'enzyme'
+import * as jestGlamorReact from 'jest-glamor-react'
+import glamorous from '../'
+
+expect.extend(jestGlamorReact.matcher)
+expect.addSnapshotSerializer(jestGlamorReact.serializer)
+
+beforeEach(() => {
+  glamorous.config.idCounter = 1
+})
+
+test('allows you to use a component as a child selector', () => {
+  const RedDiv = glamorous.div({color: 'red'})
+  const ContainerDiv = glamorous.div({
+    [RedDiv]: {
+      color: 'blue',
+    },
+  })
+
+  expect(
+    render(<ContainerDiv><RedDiv>Hello</RedDiv></ContainerDiv>),
+  ).toMatchSnapshotWithGlamor()
+})
+
+test('allows you to use a component as any kind of selector', () => {
+  const ContainerDiv = glamorous.div({
+    color: 'blue',
+  })
+  const RedDiv = glamorous.div(
+    {
+      [`${ContainerDiv} &`]: {
+        color: 'red',
+      },
+    },
+    {
+      [`${ContainerDiv} &`]: {
+        color: 'green',
+      },
+    },
+  )
+
+  expect(
+    render(<ContainerDiv><RedDiv>Hello</RedDiv></ContainerDiv>),
+  ).toMatchSnapshotWithGlamor()
+})
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -7,11 +7,17 @@ import {PropTypes} from './react-compat'
 import withTheme from './with-theme'
 import getGlamorClassName from './get-glamor-classname'
 
+let glamorousInstance = 1
+
 export default createGlamorous
 
 function createGlamorous(splitProps) {
   // TODO: in a breaking version, make this default to true
-  glamorous.config = {useDisplayNameInClassName: false}
+  glamorous.config = {
+    useDisplayNameInClassName: false,
+    idCounter: 1,
+    instanceId: glamorousInstance++,
+  }
 
   return glamorous
 
@@ -68,7 +74,14 @@ function createGlamorous(splitProps) {
           const debugClassName = glamorous.config.useDisplayNameInClassName ?
             cleanClassname(GlamorousComponent.displayName) :
             ''
-          const className = `${fullClassName} ${debugClassName}`.trim()
+          const className = [
+            fullClassName,
+            debugClassName,
+            GlamorousComponent.toStringClassName,
+          ]
+            .filter(Boolean)
+            .join(' ')
+            .trim()
 
           return React.createElement(GlamorousComponent.comp, {
             ref: props.innerRef,
@@ -94,6 +107,15 @@ function createGlamorous(splitProps) {
         })(GlamorousComponent.styles)
       }
 
+      function toString() {
+        if (!GlamorousComponent.toStringClassName) {
+          GlamorousComponent.toStringClassName = cleanClassname(
+            `g-${GlamorousComponent.glamorousId}`,
+          )
+        }
+        return `.${GlamorousComponent.toStringClassName}`
+      }
+
       Object.assign(
         GlamorousComponent,
         getGlamorousComponentMetadata({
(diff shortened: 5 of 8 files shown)


bhargav175:
Hey, I would like to help. Let me know what, why and how. I am familiar with the use case and I put my library on hold, because having this feature would be **killer**!

kentcdodds:
Cool! If you can help us figure out a good way to get a unique _and_ deterministic name for each component, that'd be awesome. You can steal ideas from styled-components. See the tweet thread I linked to.

kentcdodds:
This is ready for review!

package.json
```diff
@@ -96,12 +96,13 @@
     "rules": {
       "import/prefer-default-export": 0,
       "react/no-unused-prop-types": 0,
-      "valid-jsdoc": 0
+      "valid-jsdoc": 0,
+      "max-len": 0
```
kentcdodds: This rule is annoying when you have prettier already.

src/__tests__/__snapshots__/index.to-string-class-name.js.snap
```diff
@@ -0,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`allows you to use a component as a child selector 1`] = `
+.css-1o8fywl .g-1-239,
+[data-css-1o8fywl] .g-1-239 {
```
kentcdodds: I wonder if there's an easy way to reset this between tests 🤔  Otherwise, this snapshot's going to break anytime you add/remove glamorous components anywhere in the codebase 😢

README.md
```diff
@@ -1003,6 +1003,43 @@ const MyResponsiveDiv = glamorous.div({
...
+const Link = glamorous.a({
+  color: 'inherit',
+  textDecoration: 'none',
+  '&:hover, &:active': {
+    textDecoration: 'underline',
+  }
+})
+
+const P = glamorous.p({
+  [`& ${Link}`]: {
```
kentcdodds: Unfortunately we cannot support:

```js
const P = glamorous.p({
  [Link]: {}
})
```

Because that would be like:

```javascript
const P = glamorous.p({
  '.g-1-1': {},
})
```

Which is invalid glamor syntax :-(

src/create-glamorous.js
```diff
@@ -7,11 +7,17 @@ import {PropTypes} from './react-compat'
...
+
 export default createGlamorous
 
 function createGlamorous(splitProps) {
   // TODO: in a breaking version, make this default to true
-  glamorous.config = {useDisplayNameInClassName: false}
+  glamorous.config = {
+    useDisplayNameInClassName: false,
+    idCounter: 1,
+    instanceId: glamorousInstance++,
```
kentcdodds: Most apps will only have one instance.

src/__tests__/__snapshots__/index.to-string-class-name.js.snap
kentcdodds: Fixed this by adding a `beforeEach` to the tests.

src/__tests__/index.to-string-class-name.js
```diff
@@ -0,0 +1,47 @@
...
+import React from 'react'
+import {render} from 'enzyme'
+import * as jestGlamorReact from 'jest-glamor-react'
+import glamorous from '../'
+
+expect.extend(jestGlamorReact.matcher)
+expect.addSnapshotSerializer(jestGlamorReact.serializer)
+
+beforeEach(() => {
+  glamorous.config.idCounter = 1
```
kentcdodds: This is the best way I could think of to avoid the glamorous component IDs from growing from every other test.

README.md
paulmolluzzo: Isn't the `&` nesting implicit? Can we assume it by default when only the component name is supplied?

paulmolluzzo:
Plan to give this a more thorough review tomorrow. Catching 🎇🎆 tonight. 😀

README.md
kentcdodds: Unfortunately it cannot be :-( Because if it were then we couldn't support parent selectors:

```javascript
glamorous.button({
  [`${ButtonBar} &`]: {}
})
```

If we decide where to put the `&` you have much less power :-(

paulmolluzzo: Not even something like this?

```js
/&/.test(componentName) ? className : `& ${componentName}`
```

kentcdodds: Hmmm... Where are you suggesting that code goes? In the `toString` method, we have no insight into how the returned value will be used so we can't determine whether to include the `&` or not :-(

paulmolluzzo: Ah, yes, that's where I was thinking - in `toString()`. 😕 Oh well...

kentcdodds: It really is unfortunate... We _could_ make a glamor plugin though! That would work! If there's a selector that starts with `.class-name` we'll insert an `&` in front of it. That would work :+1:

Give me a bit of time to update for that.

kentcdodds: Alrighty! I've added a glamor plugin to handle this and I think it's pretty solid :+1:

src/glamor-plugin-nested-class-name.js
```diff
@@ -0,0 +1,41 @@
+const regex = /\.g-\d+-\d+$/
+export default nestedClassNamePlugin
+
+// this enables part of the component as selector API:
```
kentcdodds: Would love thoughts/review from @threepointone!

kentcdodds:
> What would the selector equate to if we didn't append the ampersand? 

I explain that in a big comment over the glamor plugin. Is that comment unclear? How can it be improved?

paulmolluzzo:
> I explain that in a big comment over the glamor plugin. Is that comment unclear? How can it be improved?

Ah, I was missing something about the necessity of the `&` to indicate the current element and was under the impression it was required to included the `&` to chain selectors, but it's optional for `glamor`.  See [combined selectors](https://github.com/threepointone/glamor/blob/master/docs/howto.md#combined-selectors)

My bad!

src/create-glamorous.js
taion: you should heavily document good SSR patterns here; it _can_ work, but requires the server to be careful with managing glamorous instances

README.md
```diff
@@ -1003,6 +1003,48 @@ const MyResponsiveDiv = glamorous.div({
...
+styles of other `glamorous` components (this works because they each implement
+the `toString` method):
+
+<details>
+<summary>Contextual Overrides</summary>
+
+This API is inspired by `styled-components`.
+
+```javascript
+const Link = glamorous.a({
```
taion: consider making this opt-in

src/create-glamorous.js
kentcdodds: Good point. There should probably be a rehydration strategy for the glamorous instance id and the component id... There are _very_ few cases where there will be more than one instance of glamorous (you'd have to create your own which very few (no?) people do). But the glamorous component ID should probably be rehydrated properly... 🤔 

I have no idea how to do this. Or if I'm even understanding the problem properly, or even if it is indeed a problem (which I think it is).

README.md
kentcdodds: Why do you say that?

taion: Being able to use components (viz classes) as selectors is very useful, but only very occasionally. In general, good encapsulation/modularity means you don't want to use this pattern very often – but it's very useful in the small set of cases where it's useful as all.

Speaking for myself, if I were to build such an API, I would not want components to be usable as selectors by default; I want to increase (a bit) the friction of doing this, as a hint to users that this pattern is one that should be used when necessary, rather than casually.

Or another way to put it is that most componentized styled components probably shouldn't accept arbitrary style overrides by default.

src/create-glamorous.js
taion: The core issue is that, with server rendering, different requests would use the same ID sequence for these component IDs, while each client would start from scratch.

Two concrete examples where this can cause problems:

- You create styled components dynamically. With the same `glamorous` instance, on the server, they get increased IDs every time. On the client, they don't.
- You use code splitting. The server (eventually) loads up all the bundles, but the client only loads the bundles specific to it. Actually, this is even harder to deal with, because even if you could have the "correct" bundles loaded on the server, you'd need to ensure the client loads and evaluates the static bits in the same order.

Without digging too deeply into this code, I think it'd be relatively challenging to nicely rehydrate the component IDs. The core problem is that you don't have anything to uniquely identify the components – that's why you need that ID in the first place.

Part of the problem is that, with the current implementation, the IDs get generated statically (i.e. at module import time). You could imagine an approach where you did some sort of "replay" at the time you do the server render – for example, record the sequence of IDs that are generated, and send that to the client such that the client can "replay" that sequence of IDs, but at that point it's too late; you can't observe the order in which the components are statically generated.

If you had a different instance of glamorous for each server rendering request, then you'd avoid many of these problems (assuming you could deal with the "order of split-out bundles" if using code splitting), but I guess the API here doesn't naturally let you put this on context like people do with Flux. Zones/CLS would handle that nicely, except for how they don't really exist.

README.md
kentcdodds: Ah, gotcha. I'm on the fence here. Mostly because you don't have to do anything extra with styled-components and I don't want it to be any harder to do this with glamorous than it is with styled-components (which is why I added the plugin). If we wanted it to be harder, then I could just remove the plugin and you'd have to do: `glamorous.div({ [`& ${RedDiv}`]: {/*styles*/}})` which would be slightly annoying 🤔

kentcdodds:
@taion,

Based on your comments (particularly around making this "less-optimal-pattern" opt-in and the challenges around SSR) I'm starting to wonder if we shouldn't support this and instead document this as a solution:

```javascript
const redDivClassName = 'my-own-unique-class-name-i-made-up'
const RedDiv = glamorous.div(redDivClassName, {color: 'red'})
RedDiv.className = redDivClassName

// elsewhere...
const ContainerDiv = glamorous.div({
  [`& ${RedDiv.className}`]: {
    color: 'blue',
  }
})

<ContainerDiv><RedDiv>Hello</RedDiv></ContainerDiv>

// Renders <div><div>Hello</div></div>
// with "color: blue" applied to <div>Hello</div>
```

I think that given the fact that this pattern is:
1. Not something we want to recommend people use
2. Pretty straightforward to work around/document
3. Really hard to do SSR/code splitting with

I think that maybe we should just go with document an example solution here... Thoughts?

taion:
Took a bit to check how styled-components does it: https://github.com/styled-components/styled-components/blob/v2.1.1/src/models/StyledComponent.js#L32 (Their codebase is a little bit harder to jump into, so I didn't have a chance to look until now)

Similar approach at a high level, but the identifiers are tracked per-display-name. Since in practice display name collisions are not going to be super common, it'd then be relatively hard to run into problems. Might be practical here as well?

-----

Otherwise, yeah, some variation of your pattern above looks pretty good, and probably has less scope to be surprising. It's exactly the pattern we follow in our CSS Modules code (and we name those classes like `.__button-hook` to make them more explicit).

Thinking through this some more – while this "contextual selector" support is useful, it should be sufficiently seldom-used that jumping through hoops is no big deal. For context, we use this kind of pattern like a dozen or two times across ~50k LOC, so if this is typical, then not having super smooth syntax here isn't going to really be a problem for anyone.

-----

Yeah, on net I think the example you have above is pretty good. Might be worthwhile to get feedback from @schnerd given that he was the one looking for this sort of feature in the first place.

schnerd:
The `glamorous.div({[RedDiv]: {}})` syntax is nice, but one minor concern is that it assumes I have the ability to import RedDiv. What if RedDiv is an internal of some re-usable component published on npm? If RedDiv is not exported publicly, I can't target it. In this scenario the className-based syntax is more flexible, because I might be able to use `glamorous.div({'& .RedDiv': {}})`

The className syntax certainly feels less smooth and sort of feels like we're going against css-in-js principles (that's sort of what I was trying to solve with the `$` syntax in #202), but I don't think it's prohibitively more difficult than the component-as-selector API. At the end of the day I don't have a super strong opinion between the two.

kentcdodds:
Thanks for the input @schnerd!

This is supported today: `glamorous.div({'& .RedDiv': {}})` and it seems like a pretty ok way to enable this. As an author of a reusable component library you'd either document what class names to use for style overrides or expose a way to import class names which you could use in your styles objects.

What I think would be good is for someone to create an example of how to do this with glamorous. I think that it's not a use case we want to encourage, but it's something people need to do occasionally so let's make sure people know how to do it.

braposo:
As much as I like and understand the idea of having components as selectors, it feels to me like a step back from the basic concept of using component-based styling. I see it in the same level as using context to pass down props to child components instead of simply using props.

The component as selector really feels like a good idea, but I think it's mostly because of the habit of writing CSS like that since the beginnings of time. With component-based styling, I don't think you should really use this method to style a component when you can easily pass down the props for that through `css` or even `style`. It also breaks one of the reasons why component-based styling is so popular: keeping the styles together in the component. AFAIK with component as selector it's really hard to debug styling and trace it back to the component that is actually changing the styles.

Might be a super cool feature and make it much easier for people transitioning from traditional CSS to a component styling library, but in the long run I think the experience will be worse for everyone. For that reason I agree with https://github.com/paypal/glamorous/pull/183#issuecomment-313196808 and support just documenting it and explaining why it should/shouldn't be used.

Just my 2c anyway!

kentcdodds:
Agreed. Having an escape hatch is important, but making it easier while making things more complex is not worth it I think.

kentcdodds:
Closing in favor of: https://github.com/kentcdodds/glamorous-website/issues/222

README.md
renatorib: I agree with @kentcdodds, this is annoying 'by default'.

PR #184, opened 2017-06-14
chore(theme): swap custom theme code with `withTheme`

kentcdodds:
**What**: This changes the GlamorousComponent from a class to a function that uses `withTheme`.

**Why**: This simplifies the codebase and prepares us for potentially dropping our own theming implementation in favor of the `theming` module.

**How**: Refactoring `withTheme` to have all the features that the glamorous component had and wrapping the glamorous component in `withTheme`.

This change _should_ be totally non-breaking despite some of the test changes. I'd really appreciate a review on this one from folks.

### DIFF
--- a/dist-test/index.js
+++ b/dist-test/index.js
@@ -129,7 +129,7 @@ function hasNamedExports(thing) {
   return keys.every(
     key =>
       typeof thing[key] === GlamorousComponents[key] &&
-      thing[key].name === 'GlamorousComponent',
+      thing[key].isGlamorousComponent,
   )
 }
 
--- a/package.json
+++ b/package.json
@@ -94,7 +94,8 @@
     ],
     "rules": {
       "import/prefer-default-export": 0,
-      "react/no-unused-prop-types": 0
+      "react/no-unused-prop-types": 0,
+      "valid-jsdoc": 0
     }
   },
   "lint-staged": {
--- a/src/__tests__/index.js
+++ b/src/__tests__/index.js
@@ -4,20 +4,17 @@ import * as glamor from 'glamor'
 import {render, mount} from 'enzyme'
 import * as jestGlamorReact from 'jest-glamor-react'
 import {oneLine} from 'common-tags'
-import glamorous from '../'
+import glamorous, {ThemeProvider} from '../'
 import {PropTypes} from '../react-compat'
-
 import {CHANNEL} from '../constants'
 
 expect.extend(jestGlamorReact.matcher)
 expect.addSnapshotSerializer(jestGlamorReact.serializer)
 
-const getMockedContext = unsubscribe => ({
-  [CHANNEL]: {
-    getState: () => {},
-    setState: () => {},
-    subscribe: () => unsubscribe,
-  },
+const nodeEnv = process.env.NODE_ENV
+
+afterEach(() => {
+  process.env.NODE_ENV = nodeEnv
 })
 
 test('sanity test', () => {
@@ -340,6 +337,7 @@ test('renders a component with theme properties', () => {
 
 test('in development mode the theme is frozen and cannot be changed', () => {
   expect.assertions(1)
+  process.env.NODE_ENV = 'development'
   const Comp = glamorous.div(
     {
       color: 'red',
@@ -386,22 +384,15 @@ test('passes an updated theme when theme prop changes', () => {
   expect(wrapper).toMatchSnapshotWithGlamor(`with theme prop of padding 20px`)
 })
 
-test('cleans up theme subscription when unmounts', () => {
-  const unsubscribe = jest.fn()
-  const context = getMockedContext(unsubscribe)
+test('passes `theme` to the css prop if it is a function', () => {
   const Comp = glamorous.div()
-  const wrapper = mount(<Comp />, {context})
-  wrapper.unmount()
-  expect(unsubscribe).toHaveBeenCalled()
-})
-
-test('ignores context if a theme props is passed', () => {
-  const unsubscribe = jest.fn()
-  const context = getMockedContext(unsubscribe)
-  const Comp = glamorous.div()
-  const wrapper = mount(<Comp theme={{}} />, {context})
-  wrapper.unmount()
-  expect(unsubscribe).toHaveBeenCalledTimes(0)
+  const css = jest.fn()
+  const props = {css}
+  const theme = {color: 'blue'}
+  mount(<ThemeProvider theme={theme}><Comp {...props} /></ThemeProvider>)
+  expect(css).toHaveBeenCalledTimes(1)
+  const context = expect.objectContaining({[CHANNEL]: expect.any(Object)})
+  expect(css).toHaveBeenCalledWith({...props, theme}, theme, context)
 })
 
 test('allows you to pass custom props that are allowed', () => {
@@ -431,7 +422,7 @@ test('allows you to pass custom props that are allowed', () => {
   )
 })
 
-test('should recieve inner ref if specified', () => {
+test('should receive inner ref if specified', () => {
   const getRef = jest.fn()
   const Comp = glamorous.div({
     marginLeft: '24px',
@@ -468,17 +459,13 @@ test('can accept functions which return string class names', () => {
 test('should accept user defined contextTypes', () => {
   const dynamicStyles = jest.fn()
   const Child = glamorous.div(dynamicStyles)
-  Child.contextTypes = {
-    fontSize: PropTypes.number,
-  }
+  Child.contextTypes = {fontSize: PropTypes.number}
 
-  const context = {
-    fontSize: 24,
-  }
+  const context = {fontSize: 24}
 
   render(<Child />, {context})
   expect(dynamicStyles).toHaveBeenCalledTimes(1)
   const theme = {}
-  const props = {}
+  const props = {theme}
   expect(dynamicStyles).toHaveBeenCalledWith(props, theme, context)
 })
--- a/src/__tests__/with-theme.js
+++ b/src/__tests__/with-theme.js
@@ -96,7 +96,7 @@ test('pass through when no theme provider found up tree', () => {
   expect(console.warn).toHaveBeenCalledTimes(1)
   expect(console.warn).toHaveBeenCalledWith(
     // eslint-disable-next-line max-len
-    `glamorous warning: Expected component called "Stateless Function" which uses withTheme to be within a ThemeProvider but none was found.`,
+    `glamorous warning: Expected component called "FunctionComponent" which uses withTheme to be within a ThemeProvider but none was found.`,
   )
   console.warn = originalWarn
 })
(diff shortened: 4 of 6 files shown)


dist-test/index.js
```diff
@@ -129,7 +129,7 @@ function hasNamedExports(thing) {
   return keys.every(
     key =>
       typeof thing[key] === GlamorousComponents[key] &&
-      thing[key].name === 'GlamorousComponent',
+      thing[key].isGlamorousComponent,
```
kentcdodds: Added this to make it a little more explicit.

package.json
```diff
@@ -94,7 +94,8 @@
     ],
     "rules": {
       "import/prefer-default-export": 0,
-      "react/no-unused-prop-types": 0
+      "react/no-unused-prop-types": 0,
+      "valid-jsdoc": 0
```
kentcdodds: I'm starting to love stylistic ESLint rules less and less.

src/__tests__/index.js
```diff
@@ -4,20 +4,17 @@ import * as glamor from 'glamor'
...
-import glamorous from '../'
+import glamorous, {ThemeProvider} from '../'
 import {PropTypes} from '../react-compat'
-
 import {CHANNEL} from '../constants'
 
 expect.extend(jestGlamorReact.matcher)
 expect.addSnapshotSerializer(jestGlamorReact.serializer)
 
-const getMockedContext = unsubscribe => ({
```
kentcdodds: Moved

src/__tests__/index.js
```diff
@@ -4,20 +4,17 @@ import * as glamor from 'glamor'
...
-  [CHANNEL]: {
-    getState: () => {},
-    setState: () => {},
-    subscribe: () => unsubscribe,
-  },
+const nodeEnv = process.env.NODE_ENV
+
+afterEach(() => {
+  process.env.NODE_ENV = nodeEnv
```
kentcdodds: So we don't need to worry about cleaning up after ourselves.

src/__tests__/index.js
```diff
@@ -386,22 +384,15 @@ test('passes an updated theme when theme prop changes', () => {
   expect(wrapper).toMatchSnapshotWithGlamor(`with theme prop of padding 20px`)
 })
 
-test('cleans up theme subscription when unmounts', () => {
```
kentcdodds: Moved

src/__tests__/index.js
```diff
@@ -386,22 +384,15 @@ test('passes an updated theme when theme prop changes', () => {
...
-  const unsubscribe = jest.fn()
-  const context = getMockedContext(unsubscribe)
+test('passes `theme` to the css prop if it is a function', () => {
   const Comp = glamorous.div()
-  const wrapper = mount(<Comp />, {context})
-  wrapper.unmount()
-  expect(unsubscribe).toHaveBeenCalled()
-})
-
-test('ignores context if a theme props is passed', () => {
```
kentcdodds: Moved

src/__tests__/index.js
```diff
@@ -386,22 +384,15 @@ test('passes an updated theme when theme prop changes', () => {
   expect(wrapper).toMatchSnapshotWithGlamor(`with theme prop of padding 20px`)
 })
 
-test('cleans up theme subscription when unmounts', () => {
-  const unsubscribe = jest.fn()
-  const context = getMockedContext(unsubscribe)
+test('passes `theme` to the css prop if it is a function', () => {
```
kentcdodds: Added (sorry the git diff makes this a little confusing).

Incidentally, I added this test because I thought I could get away with only wrapping `GlamorousComponent` with `withTheme` if any of the `...styles` was a function. So get a little bit of an optimization by not wrapping it for plain old objects. Turns out I can't do that because the `css` prop can be a function which accepts the `theme` and you can't know that ahead of time. So all components need to keep track of the theme. #sigh

src/__tests__/index.js
```diff
@@ -468,17 +459,13 @@ test('can accept functions which return string class names', () => {
...
-  const context = {
-    fontSize: 24,
-  }
+  const context = {fontSize: 24}
 
   render(<Child />, {context})
   expect(dynamicStyles).toHaveBeenCalledTimes(1)
   const theme = {}
-  const props = {}
+  const props = {theme}
```
kentcdodds: Here's a change. `theme` will now always be a part of the `props` on a component. I don't see that as a problem... If the user is using the `theme` prop, we have special behavior for that already (and that behavior remains the same).

In fact, in a future breaking change, I would like to remove the `theme` as the second argument and just make it part of the props, so instead of:

```javascript
glamorous.div((props, theme) => theme.someThing)
```

You could just do:

```javascript
glamorous.div(({theme}) => theme.someThing)
```

src/__tests__/with-theme.js
```diff
@@ -96,7 +96,7 @@ test('pass through when no theme provider found up tree', () => {
   expect(console.warn).toHaveBeenCalledTimes(1)
   expect(console.warn).toHaveBeenCalledWith(
     // eslint-disable-next-line max-len
-    `glamorous warning: Expected component called "Stateless Function" which uses withTheme to be within a ThemeProvider but none was found.`,
+    `glamorous warning: Expected component called "FunctionComponent" which uses withTheme to be within a ThemeProvider but none was found.`,
```
kentcdodds: FunctionComponents are not necessarily Stateless ;-)

src/__tests__/with-theme.js
```diff
@@ -129,3 +129,12 @@ test('unsubscribes from theme updates on unmount', () => {
   wrapper.unmount()
   expect(unsubscribe).toHaveBeenCalled()
 })
+
+test('ignores context if a theme props is passed', () => {
```
kentcdodds: Moved from `index.js`

src/create-glamorous.js
```diff
@@ -42,56 +42,19 @@ function createGlamorous(splitProps) {
        * This is a component which will render the comp (closure)
        * with the glamorous styles (closure). Forwards any valid
        * props to the underlying component.
-       * @param {Object} theme the theme object
-       * @return {ReactElement} React.createElement
        */
-      class GlamorousComponent extends Component {
```
kentcdodds: Able to remove quite a bit of code thanks to this change. Though we're now using the `withTheme` HoC so I'm pretty sure size will remain about the same.

src/create-glamorous.js
```diff
@@ -42,56 +42,19 @@ function createGlamorous(splitProps) {
...
+        (props, context) => {
+          /* eslint no-use-before-define: 0 */
           const {toForward, cssOverrides, cssProp} = splitProps(
             props,
             GlamorousComponent,
           )
 
           // freeze the theme object in dev mode
           const theme = process.env.NODE_ENV === 'production' ?
-            this.state.theme :
```
kentcdodds: We're now a function component so `this` doesn't work and we don't have `state`. `theme` comes from props.

src/create-glamorous.js
```diff
@@ -112,43 +75,18 @@ function createGlamorous(splitProps) {
             ...toForward,
             className,
           })
-        }
-      }
+        },
+        {noWarn: true, createElement: false},
```
kentcdodds: Brief explanation:
- `noWarn`: If there's no theme context, don't log to the console. Logging was the default behavior (because why would you use `withTheme` if there was no theme in the context?) However in our case you can totally use glamorous components without the theme context so we don't want to warn about that. So we opt-out. Default behavior remains the same (no breaking change).
- `createElement`: This is where things get a little tricky. In our case, `GlamorousComponent` is now actually just a regular function that's called by the `ThemedComponent`. This way we avoid adding another layer in the React Tree and we don't actually need lifecycle hooks for this component anyway so we're covered. Default case is `true` so that is unchanged (no breaking change).

src/create-glamorous.js
```diff
@@ -112,43 +75,18 @@ function createGlamorous(splitProps) {
...
-      GlamorousComponent.propTypes = {
+      const glamorousPropTypes = {
         className: PropTypes.string,
         cssOverrides: PropTypes.object,
         theme: PropTypes.object,
         innerRef: PropTypes.func,
         glam: PropTypes.object,
       }
-
-      const defaultContextTypes = {
```
kentcdodds: Moved

src/create-glamorous.js
```diff
@@ -166,7 +104,7 @@ function createGlamorous(splitProps) {
           forwardProps,
           displayName,
         }),
-        {withComponent},
+        {withComponent, isGlamorousComponent: true},
```
kentcdodds: Just thought this would be handy :)

src/with-theme.js
```diff
@@ -44,13 +59,45 @@ export default function withTheme(ComponentToTheme) {
...
+      }
     }
   }
 
-  ThemedComponent.contextTypes = {
+  const defaultContextTypes = {
     [CHANNEL]: PropTypes.object,
   }
 
+  let userDefinedContextTypes = null
```
kentcdodds: Moved from `GlamorousComponent`

src/create-glamorous.js
```diff
@@ -123,33 +87,6 @@ function createGlamorous(splitProps) {
         glam: PropTypes.object,
       }
 
-      const defaultContextTypes = {
```
kentcdodds: Moved

src/with-theme.js
```diff
@@ -3,37 +3,52 @@ import React, {Component} from 'react'
 import {CHANNEL} from './constants'
```
kentcdodds: The changes in this file are mostly to refactor it to be more in line with what we had for regular GlamorousComponents. I believe that we always intended this to work the same way that the GlamorousComponent implementation worked and it didn't in some subtle ways. Good to bring these into alignment.

kentcdodds:
I've published this as `glamorous@beta` if anyone would like to take it for a spin and make sure it doesn't bust your use case.

src/__tests__/index.js
kwelch: This would also move `context` to second arg which matches stateless components. 👍

kwelch:
I will pull this into a small project I have that leverages themes tomorrow morning.

kentcdodds:
Thanks @kwelch! :tada:

jfrolich:
Is there any API change?

kwelch:
I just switched to the `beta` in a small app I have that uses theme and saw no issues and no api changes, just a simple `yarn add glamorous@beta` 🎉

kwelch:
I just switched to the `beta` in a small app I have that uses theme and saw no issues and no api changes, just a simple `yarn add glamorous@beta` 🎉

kentcdodds:
@jfrolich, no API change. But it's a bit of a refactor in code I didn't write, so I wanted to be sure 😀

Sounds like we're good to go!

iamstarkov:
🎉

jfrolich:
Let me test it on our codebase as well, we are extensively using theming

Issue #177, opened 2017-06-12
Add flow type definitions


kentcdodds:
Because TypeScript shouldn't have all the fun :wink:

I'm pretty new to Flow. Anyone with experience wanna help out? I think having the existing TypeScript definitions could really help :) And in our case we could probably add the flow definitions directly in the source because that would be really handy as well and we're already stripping them out with the React preset :+1: Though, I'm not sure whether consumers would be able to use those. Maybe those are two separate issues (adding flow to the codebase and adding type definitions for consumers to use?) I don't know! Help wanted!

ErikCupal:
Sounds great! I'll definitely help. I have not tried Flow yet, but it's 90% the same as Typescript as far as I know.

kentcdodds:
Sounds awesome!

kentcdodds:
Thanks! :)

pbomb:
I would love to see this and could probably help out as well. I think [this post](http://javascriptplayground.com/blog/2017/01/npm-flowjs-javascript/) did a good job advocating for the approach of running the code through babel + creating `.js.flow` files next to the source files so that consumers would get the benefits of flow types.

kentcdodds:
Fantastic! Looks really great. Thanks for ya'lls willingness to pitch in :)

suchipi:
I'm also interested in helping; is there a plan for dividing up the codebase so we don't duplicate work and step on each other's toes? Or would it be better if only one person worked on this?

luke-john:
I'd be keen for the typescript and flow typings to follow a similar structure.

This should help keep the two in sync and make the initial setup of flow typing a bit easier.

Happy for the typescript typing to get restructured as part of this if there's a pattern more suitable.

torifat:
@kentcdodds How about this one? https://babeljs.io/docs/plugins/transform-flow-comments/

kentcdodds:
I think that @pbomb's suggest would support consumers as well. But if this works, it seems simpler! Anyone wanna try your hand at a PR for that? @suchipi, you seemed interested. Would you like to? Whoever wants to, just mention here that you've started working on it when you start

suchipi:
Yeah, I'm interested. I will take a look this evening, using babel-plugin-transform-flow-comments.

suchipi:
[...]

PR #205, opened 2017-07-02
WIP: Add flow type definitions

suchipi:
This is still a work in progress (close to done), but I'd like to get some more eyes on it because I feel like I'm working in a vacuum.

**What**: Adds flow type definitions to the src

**Why**: #177

**How**:
* Added flow-bin to the project and configured the project for flow
* Added flow types to the files in src

**Checklist**:

- [ ] Documentation
- [ ] Tests N/A
- [ ] Code complete

TODO:
- [x] Add types to constants.js
- [x] Add types to create-glamorous.js
- [x] Add types to constants.js
- [x] Add types to dom-elements.js
- [x] Add types to get-glamor-classname.js
- [ ] Add types to index.js
- [ ] Add types to react-compat.js (We can probably skip this one)
- [ ] Add types to react-props.js (We can probably skip this one)
- [ ] Add types to should-forward-property.js (We can probably skip this one)
- [x] Add types to split-props.js
- [x] Add types to theme-provider.js
- [x] Add types to tiny.js
- [x] Add types to umd-entry.js (Can we skip this one?)
- [x] Add types to with-theme.js
- [ ] Verify that the type signatures in create-glamorous.js are correct
- [ ] Instead of stripping the flow types (which the react preset does), export them as comments
- [ ] Figure out what to do about verification failures (coverage not 100%, size > 5kb, etc)
- [ ] Submit the library types to flow-typed


suchipi:
@kentcdodds Can you review this (particularly `create-glamorous.js`) and make sure the types look sane? About half of them I had to guess on because the documentation in the README was somewhat vague about every possible allowed format.

src/create-glamorous.js
```diff
@@ -37,40 +73,46 @@ function createGlamorous(splitProps) {
...
      */
-    function glamorousComponentFactory(...styles) {
+    function glamorousComponentFactory(...styles: Array<Style>) {
       /**
        * This is a component which will render the comp (closure)
        * with the glamorous styles (closure). Forwards any valid
        * props to the underlying component.
        */
-      const GlamorousComponent = withTheme(
+      let FinalGlamorousComponent // eslint-disable-line prefer-const
```
kentcdodds: Is there a reason that this needs to be `let`? It isn't being reassigned so we can make it `const`.

src/types/Style.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import type {Theme} from '../theme-provider'
+import type {CSSProperties} from './CSSProperties'
+
+type StyleFunction = (
+  {[key: string]: any}, // props
```
kentcdodds: One key we _know_ this will always have is the `theme` key which is the `Theme` type.

src/types/Style.js
```diff
@@ -0,0 +1,17 @@
...
+  Theme, // theme
+  ?{[key: string]: any}, // context
+) => string | CSSProperties;
+
+// TODO: Is null okay here?
```
kentcdodds: Yes, `null` is fine

src/split-props.js
```diff
@@ -1,4 +1,35 @@
...
+  RefCallback,
+} from './types/React'
+import type {CSSProperties} from './types/CSSProperties'
+import type {Theme} from './theme-provider'
+
+export type SplittableProps = {
+  css: CSSProperties,
+  theme: Theme,
+  className: string,
+  // TODO: Use rootEl to give a more specific typing for this innerRef's
```
kentcdodds: Not sure what this means...

src/theme-provider.js
```diff
@@ -1,7 +1,32 @@
+// @flow
 import React, {Component} from 'react'
 import brcast from 'brcast'
+import type {Element} from 'react' // eslint-disable-line no-duplicate-imports
+
+// eslint-disable-next-line no-duplicate-imports
```
kentcdodds: If we can't combine the imports, then let's just turn off this rule.

src/theme-provider.js
```diff
@@ -1,7 +1,32 @@
...
+type Props = {
+  theme: Theme,
+  children: string | number | false | null | Array<Element<*>>,
+};
+
+// TODO: {[typeof CHANNEL]: void} is an object with the entry
+// `[CHANNEL]: undefined`, but an empty object should be allowed here, too.
+// Flow complains about the missing property if I type it as {}, though,
+// even though an empty object should still return `undefined` when
+// accessing the missing property.
```
kentcdodds: Hmmm... I'm unsure how to address this.

src/split-props.js
```diff
@@ -1,4 +1,35 @@
...
+import type {
+  FunctionalComponent,
+  ComponentClass,
+  RefCallback,
+} from './types/React'
+import type {CSSProperties} from './types/CSSProperties'
+import type {Theme} from './theme-provider'
+
+export type SplittableProps = {
+  css: CSSProperties,
```
kentcdodds: This is actually a `Style` :)

src/split-props.js
```diff
@@ -1,4 +1,35 @@
...
+
+export type SplitPropsOptions = {
+  propsAreCssOverrides?: boolean,
+  rootEl: string | FunctionalComponent | ComponentClass,
+  forwardProps: Array<string>,
+};
+
+export type SplitPropsResult = {
+  toForward: {[key: string]: any},
+  cssProp: CSSProperties,
```
kentcdodds: This is a `Style`

src/with-theme.js
```diff
@@ -66,8 +91,11 @@ export default function withTheme(
         // as our `render` method without going through lifecycle hooks.
         // Also allows us to forward the context in the scenario where
         // a user wants to add more context.
-        // eslint-disable-next-line babel/new-cap
-        return ComponentToTheme({...this.props, ...this.state}, this.context)
+
+        /* eslint-disable max-len */
+        // $FlowFixMe: This may produce undesirable behavior when ComponentToTheme is a class instead of a function.
```
kentcdodds: Correct, but this is not technically a public API (at least, it wont be supported). It's only used for creating glamorous components. Didn't we discuss somewhere that we could make things work by extending `ThemedComponent` instead of doing this weird stuff though? I'll give it a look.

kentcdodds:
I gotta be honest with you, as excited as I am about flow and typed JavaScript, this PR kinda scares me a little bit because I'm pretty inexperienced and I'm worried that I wont be able to maintain it very well 😨

kentcdodds:
This doesn't mean that I definitely wont merge it. But it is giving me pause to think...

src/create-glamorous.js
suchipi: It actually is reassigned, on line 152. I'm not sure why eslint wasn't detecting this properly.

We need a binding that will eventually contain the class because the component we pass into `withTheme` needs access to it. So this part is a little bit awkward.

src/split-props.js
suchipi: Cool, I'll change it

src/split-props.js
suchipi: Ah, this isn't necessarily important, but I was saying that instead of typing innerRef as `RefCallback<Component<*, *, *>>` (aka ref callback of any component instance), if rootEl is set to eg. "div", it could be `RefCallback<HTMLDivElement>`.

Looking at it again, unless I add in some knowledge of when it'll be a component instance and when it'll be an html node, this should probably use `RefCallback<any>` instead, because otherwise the type will be incorrect when the ref is for an html element.

src/theme-provider.js
suchipi: This is actually the same issue I run into [here](https://github.com/paypal/glamorous/pull/205/files/6526214b9f057d26db60dd3856b75593c0afa0af#diff-3be2a7cdd58ade04e2acbcef3702d3d7R195).

The issue is that flow doesn't refine a value into a more specific type based on a property value unless both types have the same property.

So this works:
```js
type Success = { error: false, data: any }
type Failure = { error: true, errorMessage: string }

function handleResponse(response: Success | Failure) {
  if (!response.error) {
    // Flow knows that this must be a Success, so it lets you
    // access data.
    return response.data
  }
}
```
but this does not:
```js
type Success = { error: false, data: any }
type Failure = { error: true, errorMessage: string }

function handleResponse(response: Success | Failure) {
  if (!response.errorMessage) {
    // Flow cannot guarantee that response is a success
    return response.data
  }
}
```

This is because object types in flow are not "exact" by default; that is, they're allowed to have extra properties:
```js
type SomethingWithValue = { value: number }

const a: SomethingWithValue = { value: 4 } // this is fine
const b: SomethingWithValue = { value: 5, foo: 6 } // this is fine, too
```

Unfortunately, even if you make your types exact, it looks like flow still can't refine them (`{| ... |}` is "exact object type" syntax):
```js
type Success = {| error: false, data: any |}
type Failure = {| error: true, errorMessage: string |}

function handleResponse(response: Success | Failure) {
  if (!response.errorMessage) {
    // Flow still can't make any guarantees
    return response.data
  }
}
```

That might be a bug, but even if flow could refine exact types based on unrelated property presence, using an exact type in this situation wouldn't be correct, since context can have anything in it.

src/with-theme.js
suchipi: That's what I did [here](https://github.com/paypal/glamorous/pull/205/files/6526214b9f057d26db60dd3856b75593c0afa0af#diff-3be2a7cdd58ade04e2acbcef3702d3d7R134). I don't think that fix would apply to this situation, unless I'm misunderstanding

suchipi:
I had the same thought... it might be better to pursue external libdefs (like what you're doing for typescript) instead of using flow in the source. Some of the patterns the code uses are not things that flow understands very well, which already makes it a bit difficult, and I don't use glamorous, so I don't think I'd be able to volunteer to maintain the flow typings in your stead.

kentcdodds:
> I had the same thought... it might be better to pursue external libdefs (like what you're doing for typescript) instead of using flow in the source.

Yeah, I think I like this better as well. Another thing I want to consider is how easy it is for anyone (experienced or not) to contribute to the codebase and I feel that while this really helps with the stability of the codebase, it does make it a little harder to contribute to.

Sorry that much of your work will not be merged, but if you could externalize it I think I'd be more comfortable with that. Thank you so much for all of this work, especially because you don't even use glamorous! I hope that you learned a lot and found this experience to be informative and fun!

luke-john:
Also as an option, flow defs can be automatically generated from typescript definitions (not the other way round though).

I'll hopefully have a chance to look at these typings tomorrow and update the typescript types for some of the currently missing things that have been added here.

I'll also throw up a spike of having flow defs generated from the typescript definitions shortly.

suchipi:
I'm going to close this for now, I may revisit this in the future (as external types instead), but if someone else wants to pick that task up, they are welcome to.

abs-zero:
What is left to be done here? Or all the work you have done here should be re-written?

suchipi:
The work should be rewritten as libdefs instead of in-src flow types

abs-zero:
I see. Maybe I will try to work on this today.

kentcdodds:
That's great! Thanks!

luke-john:
Cheers, that would be excellent.

@abs-zero the typescript typings may be of help with this.  If you do want to look at them I'd recommend checking out the ones on [this topic branch](https://github.com/paypal/glamorous/pull/230), it's got a lot more coverage and the names are more similar to the glamorous documentation.

I'll try and keep an eye on the [glamorous gitter chat](https://gitter.im/paypal/glamorous) throughout the day if you've got any questions about the typings.

Issue #134, opened 2017-05-25
Improve Server Side Rendering


kentcdodds:
The [styled-components v2 announcement blogpost](https://medium.com/styled-components/announcing-v2-f01ef3766ac2) states:

> When you server-side render an app with 💅 styled-components it will automatically only send the critical CSS down to the client.

I'm pretty sure that we don't actually do that right now. As soon as you say:

```javascript
const Button = glamorous.button({ /* styles */})
```

Those styles get added to glamor... wait...

jkjk, we totally don't add them to glamor until rendering time. Just verified this with:

```javascript
import React from 'react'
import ReactDOM from 'react-dom'
import {styleSheet} from 'glamor'
import glamorous from '../'

const Button = glamorous.button({marginLeft: 20})
console.log(styleSheet.inserted)
// -> {}
const x = <Button />
console.log(styleSheet.inserted)
// -> {}
ReactDOM.render(x, document.createElement('div'))
console.log(styleSheet.inserted)
// -> { '140tz9x': true }
```

So I think we're good. EXCEPT!

We _could_ probably improve the experience a bit. If you look at the Next.js example for server side rendering [here](https://github.com/zeit/next.js/blob/master/examples/with-glamorous) there's a bit of boilerplate, and the styled-components blogpost says:

> On top of that we support concurrent server-side rendering to avoid blocking your process with every request and we automatically rehydrate your styles in the browser.

And further:

> All it takes is four lines of code

Let's see if we can improve this experience a bit taking inspiration (again) from styled-components :)

nitin42:
I'd like to work on it (assuming this is still open).

kentcdodds:
Awesome! Yeah! Let me know what you're thinking for doing :)

nitin42:
Sure!

nitin42:
Possible solution :

Whenever an app is rendered on the server, we need to include the generated styles on the pages. This can be done by passing a component to a callback which will return html or array of React elements. Further, taking inspiration from styled-components, we could gather all the style tags and include the extracted css in the html  by creating an instance of `ServerStyleSheet` and pass array of React elements (return by the callback, eg - renderToStaticMarkup) to the body.

```html
<html>
  <head>
    {style_tags}
  </head>
  <body>
    <div className="root">
      {react_elements_to_be_render}
    </div>
  </body>
</html>
``` 

Let me know what you think (I am still not sure about this).

kentcdodds:
[...]

PR #206, opened 2017-07-03
SSR implementation

nitin42:
**What**: Server Side Rendering 

**Why**: #134 

**How**: 
* Added `renderStaticMarkup` to collect the html, css and rendered class names (returned by glamor)

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete

Still needs improvements though !


nitin42:
Because we collect all of the calls to `css()` internally using glamor, so we have to rely `renderStatic` to return the generated html and css. We may need to rewrite this from scratch if the glamorous API changes in the future.

kentcdodds:
> We may need to rewrite this from scratch if the glamorous API changes in the future.

That's fine 👍 

> Also that `glamor` provides `renderStaticOptimized`, we could add this also to the `StyleSheetServer`.

I think that we should just use `renderStaticOptimized`. I don't see any reason not to do that... 🤔

kentcdodds:
Hmmm.... I was actually refactoring your PR in the GitHub editor and this is what I ended up with:

screen shot 2017-07-04 at 2 09 29 pm

Seems pretty pointless... We're not really doing much... 🤔 I wonder if this is really just a matter of creating more examples 🤔

nitin42:
🤔  I think it does make a point because `Glamorous` is using `glamor` as a peer dependency. For SSR , we need to keep track of all the `css()` calls and since we rely on `glamor`, we can use `renderStaticOptimized` from `glamor` ? Or should we write our own method  ?

kentcdodds:
What I was suggesting is instead just improve our documentation for how to do SSR with glamor (which does now and always will work with glamorous).

nitin42:
Seems better to me. I would create some examples and then we can document it later ?

nitin42:
Good idea 👍 . I could create some examples taking inspiration from styled-components, aphrodite and glamor !

kentcdodds:
Sounds great. I'll go ahead and close this one. Thanks!

nitin42:
Hi there ! I am a little busy maintaining my project [`terminal-in-react`](https://github.com/nitin42/terminal-in-react). I will start working on the examples soon. Cool ?

kentcdodds:
Sounds good! Thanks! Maybe you could just add them directly to the new website: github.com/kentcdodds/glamorous-website

nitin42:
Sure!

PR #207, opened 2017-07-04
Update README.md

### LINKED ISSUE #198: Change `theme` docs
A while back, we changed the way theming works slightly. It's now available as part of the props in addition to the second argument to the dynamic styles function.

I think that accessing the theme via props is better than using the second argument. It's less confusing. Also, I'm hoping that in the next breaking change, we can remove the `theme` as the second argument entirely (so the second argument would be `context` which is currently the third argument).

So anywhere we do something like this:

```javascript
glamorous.div(({big}, theme) => ({margin: big ? 20 : 10, color: theme.colors.main}))
```

Should be changed to:

```javascript
glamorous.div(({big, theme}) => ({margin: big ? 20 : 10, color: theme.colors.main}))
```

It's really nice if you're only accessing the theme as well. It means instead of:

```javascript
glamorous.div((propsThatAreNotUsed, theme) => ({}))
```

You can just do:

```javascript
glamorous.div(({theme}) => ({}))
```

Thoughts?

WillowHQ:
see line 666-671 I have updated the code snippet - is this what you want ?

Attempting to close #198 

Requested from Kent as a good issue for a first time contributor to attempt. 

Editing the readme.md file. 

**Checklist**:

- [X] Documentation
- [ ] Tests "N/A"
- [ ] Code complete "N/A"
- [ ] Followed the commit message format

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ React component styling solved with an elegant ([inspired](#inspiration)) API,
 [![downloads][downloads-badge]][npm-stat]
 [![MIT License][license-badge]][LICENSE]
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-42-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-43-orange.svg?style=flat-square)](#contributors)
 [![PRs Welcome][prs-badge]][prs]
 [![Chat][chat-badge]][chat]
 [![Code of Conduct][coc-badge]][coc]
@@ -664,7 +664,7 @@ const secondaryTheme = {
 // a themed <Title> component
 const Title = glamorous.h1({
   fontSize: '10px'
-}, (props, theme) => ({
+}, ({theme}) => ({
   color: theme.main.color
 }))
 
@@ -707,7 +707,7 @@ const theme = {
 // a themed <Title> component
 const Title = glamorous.h1({
   fontSize: '10px'
-}, (props, theme) => ({
+}, ({theme}) => ({
   color: theme.main.color
 }))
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -661,10 +661,13 @@ const secondaryTheme = {
   main: {color: 'blue'}
 }
 
+// attempting to close #198
```
paulmolluzzo: This line can be removed.

README.md
```diff
@@ -661,10 +661,13 @@ const secondaryTheme = {
...
 // a themed <Title> component
 const Title = glamorous.h1({
-  fontSize: '10px'
-}, (props, theme) => ({
+  fontSize: '10px',
```
paulmolluzzo: Need to close the first object here:

```js
const Title = glamorous.h1({
  fontSize: '10px'
}, ({props, theme: { main }} => ({
```

README.md
```diff
@@ -661,10 +661,13 @@ const secondaryTheme = {
...
+// attempting to close #198
 // a themed <Title> component
 const Title = glamorous.h1({
-  fontSize: '10px'
-}, (props, theme) => ({
+  fontSize: '10px', 
+  ...props, 
+  theme: theme }
+ => ({
   color: theme.main.color
```
paulmolluzzo: Assuming the comment above this can just be `main.color`

README.md
```diff
@@ -661,10 +661,13 @@ const secondaryTheme = {
...
 
+// attempting to close #198
 // a themed <Title> component
 const Title = glamorous.h1({
-  fontSize: '10px'
-}, (props, theme) => ({
+  fontSize: '10px', 
+  ...props, 
+  theme: theme }
+ => ({
```
kentcdodds: Yeah, I think this codeblock needs to be:

```javascript
const Title = glamorous.h1({
  fontSize: '10px'
}, ({theme}) => ({
  color: theme.main.color
}))
```

paulmolluzzo: Yeah, not even using `props`. 👍

PR #231, opened 2017-07-13
Support React Primitive interfaces

nitin42:
**What**: Export for React Primitives 

**Why**: For adding the support to render the primitive interfaces across the targets

**How**: 
* Defined getter for each aliases.
* Added glamorous constructor for primitives using composition.
* Export `ThemeProvider` and `withTheme`

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged 

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
   "devDependencies": {
     "@types/react": "^15.0.24",
     "all-contributors-cli": "^4.3.0",
-    "animated": "0.2.0",
     "babel-cli": "^6.24.1",
     "babel-jest": "^20.0.3",
     "babel-plugin-external-helpers": "^6.22.0",


src/primitives/index.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// Native UI components
+const aliases = 'Image Text Touchable View'
+
+/* Lowercase the aliases according to the built-in React Native component
+factories, define the getter for each alias and then pass it to
+glamorousComponent */
+
+aliases.split(/\s+/m).forEach(alias =>
```
kentcdodds: Could we just use a regular array instead? I think that's more straightforward:

```javascript
const aliases = ['Image', 'Text', 'Touchable', 'View']
aliases.forEach(alias =>
...
```

src/primitives/index.js
```diff
@@ -0,0 +1,36 @@
...
+  Object.defineProperty(glamorousComponent, alias.toLowerCase(), {
```
kentcdodds: Is there a reason we're not simply doing:

```javascript
glamorousComponent[alias.toLowerCase()] = createComponent(reactPrimitives[alias])
```

?

src/primitives/index.js
```diff
@@ -0,0 +1,36 @@
+import reactPrimitives from 'react-primitives'
+import glamorous, {ThemeProvider, withTheme} from '../index'
+
+// Create glamorous component with composition
+function createComponent(componentConstructor, tag) {
```
kentcdodds: Does this need to accept a `componentConstructor`? Can't this simply just reference `glamorous` as that's the only use case for this function anyway?

src/__tests__/primitives.js
```diff
@@ -0,0 +1,99 @@
...
+    })
+    const ReactImage = glamorousComponent.image({padding: '10px'})
+    expect(
+      <ReactImage
+        style={styles.image}
+        source="https://facebook.github.io/react/img/logo_og.png"
+      />,
+    ).toMatchSnapshotWithGlamor()
+  })
+})
```
kentcdodds: Can we also support the jsx-style inspired api?

```javascript
<glamorous.Image
  justifyContent="center"
  alignItems="center"
  padding={10}
  source="https://facebook.github.io/react/img/logo_og.png"
/>
```

src/primitives/index.js
nitin42: So that these aliases cannot be changes or deleted from the `glamorousComponent` !

src/primitives/index.js
nitin42: Sure, we can totally do that but as we are not using any type system in the codebase so I thought it would be better if we could fix the invalid 'tag' type.

src/primitives/index.js
nitin42: Ok! I will change it.

src/__tests__/primitives.js
nitin42: May be !? I am not sure about this. I can look into this further if you want so.

src/primitives/index.js
kentcdodds: > if we could fix the invalid 'tag' type.

Could you elaborate? I don't see any harm in just making the code simpler. No need to have the extra function...

src/primitives/index.js
kentcdodds: I can't think of a scenario where that would happen. I'd rather not make the codebase more complex just to handle that use case...

src/__tests__/primitives.js
kentcdodds: Yes please. `glamorous` and `glamorous-native` both support this API and it's an important part of glamorous.

nitin42: I was analysing the snapshots and I came to know that styles in React Native are mapped to number. When I'm trying to specify the style properties on a component, it just gives a number which is reference to the styles. 🤔

kentcdodds:
Hi! What's help do you need to get this working?

nitin42:
Thanks! I'll look into this (`jsx-style` api) further and will let you know if it's working or I need some help 😄

nitin42:
I think I've got a way to add the JSX api and I will push the changes soon. But if there is any deadline  or something let me know!

nitin42:
One thing, `Text` and `View` are already assigned on `glamorous`. 

```
{ [Function: GlamorousComponent]
  propTypes: 
   { className: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     cssOverrides: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     theme: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     innerRef: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     glam: { [Function: bound checkType] isRequired: [Function: bound checkType] } },
  contextTypes: [Getter/Setter],
  styles: [],
  comp: 'text',
  rootEl: 'text',
  forwardProps: [],
  displayName: 'glamorous.Text',
  propsAreCssOverrides: true }
```
 🤔  ?

kentcdodds:
> But if there is any deadline or something let me know!

Open source has no deadlines ;-) But I am hoping to do a big push of stuff in the next week or so. Would love to include this as part of that!

package.json
```diff
@@ -44,6 +45,7 @@
   "devDependencies": {
     "@types/react": "^15.0.24",
     "all-contributors-cli": "^4.3.0",
+    "animated": "0.2.0",
```
kentcdodds: Not sure what this is here for. Let's make sure to remove it if it's unneeded.

nitin42:
Ok! So I've fixed some issues and now we can easily style the primitives with glamorous. But I noticed something that prop `View`, `Text` and `Image` are already assigned on `glamorous`.

```
{ [Function: ThemedComponent]
  propTypes: 
   { className: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     cssOverrides: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     theme: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     innerRef: { [Function: bound checkType] isRequired: [Function: bound checkType] },
     glam: { [Function: bound checkType] isRequired: [Function: bound checkType] } },
  contextTypes: [Getter/Setter],
  styles: [],
  comp: 'text',
  rootEl: 'text',
  forwardProps: [],
  displayName: 'glamorous.Text',
  withComponent: [Function: withComponent],
  isGlamorousComponent: true,
  propsAreCssOverrides: true }
```
Can you tell me something about this ? Also I tried implementing the `jsx` api for the primitives but I am getting `css-nil` as classname. I am doing this - 

```js
import reactPrimitives from 'react-primitives';
import createGlamorous from './create-glamorous'
import splitProps from './split-props'

const glamorous = createGlamorous(splitProps)
const aliases = ['Image', 'Text', 'View', 'Touchable'];

Object.assign(
  glamorous,
  aliases.reduce((getters, alias) => {
    getters[alias.toLowerCase()] = glamorous(reactPrimitives[alias])
    return getters
  }, {}),
)

Object.assign(
  glamorous,
  aliases.reduce((comps, alias) => {
    comps[alias] = glamorous[alias.toLowerCase()]()
    comps[alias].displayName = `glamorous.${alias}`
    comps[alias].propsAreCssOverrides = true
    return comps
  }, {}),
)
```

package.json
nitin42: Sorry! I will remove this ..

kentcdodds:
I think that I've missed a few comments from you here. How can we help? What questions do you have?

kentcdodds:
I think you asked about the existence of `glamorous.Text` and `glamorous.View`. Those are there because `<text>` and `<view>` are svg elements so they're supported by the default `glamorous` export. Does glamorous-primitives need to include all the html elements? I would expect it only needs to support Image, Text, Touchable, and View...

nitin42:
No we only need to add Image, Text, Touchable and View. I've a doubt!

We may need to create a native constructor for glamorous to style the primitives. According to our api when we import `glamorous` from `glamorous/primitives`, this will only render on web!

```js
const StyledView = glamorous.View({ color: 'red' });

return (
  <StyledView>
     Hello World
  </StyledView>
)
```
But we also want it to render on mobile platform with the same import!

kentcdodds:
Yeah, I'm not sure how to do that 🤔 But we definitely want it rendering anywhere you can render primitives....

nitin42:
If you want I can look into this further. Seems interesting though!

kentcdodds:
Yes, please do! Thanks for all your work so far!

kentcdodds:
Gonna close this because it's done!

PR #233, opened 2017-07-17
Add glam-props to related projects

bulby97:
What: Add glam-props to related projects
Why: Because I can :D


luke-john:
Awesome stuff, cheers.

It would be great to add this to the new websites [integration page](https://rc.glamorous.rocks/integrations/).

https://github.com/kentcdodds/glamorous-website/tree/master/pages/integrations/content

kentcdodds:
Good point @luke-john, maybe we should just focus on that and remove this section from the README entirely (once the site is published 🔜)

Do you mind adding a PR to put this in integrations on the new website @bulby97?

kentcdodds:
Added to the website

PR #234, opened 2017-07-18
add codemods

kentcdodds:
**What**: This adds a codemod. It's still a work in progress

**Why**: To help people upgrade

**How**: A lot of help and inspiration from @tdeschryver

**Checklist**:

- [x] Documentation
- [x] Tests (more needed)
- [ ] Ready to be merged 

We need more test cases! There are definitely code styles that this doesn't handle. We should also probably consider the `css` prop which can accept dynamic style functions as well.

### DIFF
--- a/other/codemods/README.md
+++ b/other/codemods/README.md
@@ -1,0 +1,52 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [glamorous codemods](#glamorous-codemods)
+  - [Running Codemods](#running-codemods)
+  - [Available codemods](#available-codemods)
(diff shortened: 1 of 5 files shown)


other/codemods/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+To
+
+```javascript
+function(props, context) {}
+```
+
+Where `props` has a `theme` property.
+
+This codemod updates your code to use the `theme` from props and remove the
```
tdeschryver: A small one, but I think this should be ... and remove`s` the ... ?

kentcdodds: Works either way. I read it as: "updates your code ... to remove the `theme` argument"

tdeschryver: Oh, OK! Now I feel stupid, sorry bothering 😅

kentcdodds: You're no bother! You really helped get this going. Thank you!

PR #239, opened 2017-07-21
Update styled applied comments

### LINKED ISSUE #238: Styles output representation in README
I saw something in some of code snippets in `README.md`. e.g. in [glamorousComponentFactory](https://github.com/paypal/glamorous#glamorouscomponentfactory) section

In this code snippet: 
``` 
const UnstyledComp = ({ className, children }) => <div className={`${className} other-class`}>{children}</div>
const MyStyledComp = glamorous(UnstyledComp)({ margin: 1 })

<MyStyledComp>content</MyStyledComp>
// rendered output: <div class="<glamor-generated-class> other-class">content</div>
// styles applied: {margin: 1}
```

This part is unclear: 
```
// styles applied: {margin: 1}
```

When we are applying styles with CSS classes and not with JS objects, why we should represent applied styles with objects? In reality the styles applied are like this `margin: 1px;` not `{margin: 1}`. This is what we are passing in to get that out.

morajabi:
Fixes #238

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,10 @@ const MyStyledComp = glamorous(UnstyledComp)({ margin: 1 })
 
 <MyStyledComp>content</MyStyledComp>
 // rendered output: <div class="<glamor-generated-class> other-class">content</div>
-// styles applied: {margin: 1}
+// styles applied: 
+// <glamor-generated-class> {
+//  margin: 1px;
+// }
 ```
 
 ##### ...styles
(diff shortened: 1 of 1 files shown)


kentcdodds:
Would you like to [add yourself](../blob/master/CONTRIBUTING.md#add-yourself-as-a-contributor) to the [contributors table](../#contributors)?

morajabi:
Thanks. Sure. I will do it in another PR.

PR #241, opened 2017-07-23
WIP: Add flow typings

### LINKED ISSUE #177: Add flow type definitions

abs-zero:
This is a work-in-progress PR. Might need several changes in accordance to reviews. Also many thanks to @luke-john for continuously helping me in gitter chat.

**What**: Adds flow types definitions

**Why**: Resolves #177 

**How**: A single file named `glamorous.js` is added inside `flow-typed` directory. I think this should be more discussed upon whether the file should be kept at [flow-typed](https://github.com/flowtype/flow-typed) 
or here. Everything is in a single file because importing types from another file in a definition file did not work appropriately. If anyone knows anything about this, please do help.

**Checklist**:

- [N/A] Documentation (I will write in new site)
- [x] Tests
- [x] Ready to be merged 

**TODO**:
- [x] Add types for glamorous function and GlamorousFactoryFunction ( e.g, `glamorous()()` )
- [x] Add types for ThemeProvider and WithTheme
- [x] Add types for all built-in component factories (e.g: `glamorous.div`, `glamorous.section`)
- [x] Add types for all built-in glmaorous components (e.g. `glamroous.Div`)
- [x] Add type for WithComponent


abs-zero:
I am having difficult time to implement themes due to potential [flow bugs](https://github.com/facebook/flow/issues/4447)

kentcdodds:
I'm _way_ excited about this! Any chance this can be merged before I do the big release of a bunch of stuff? That'd be great :smile:

Also, it might be good to have a few test files with the flow stuff like we have with the TypeScript stuff...

In addition, I'm going to change the base branch of this PR to the `next` branch so it will be released with that stuff. Could you [rebase your changes](https://egghead.io/lessons/javascript-how-to-rebase-a-git-pull-request-branch) onto the `next` branch? Thanks!

abs-zero:
Sorry for the late reply. Just done rebasing and I will complete the remaining work. 
There's one issue here though. I don't think keeping in `flow-typed` folder is way distributing types to users. So should we move typedefs to [flow-typed](https://github.com/flowtype/flow-typed) like styled-component does? I want flow experts to have an opinion on this. :smile:

kentcdodds:
I [tweeted for help](https://twitter.com/kentcdodds/status/890237240906928128). We'll see if someone knows, because I don't!

kentcdodds:
Some folks are saying flow-typed is the way to go. I'd be inclined to agree :+1:

abs-zero:
So, should I close this and make a new PR in flow-typed? Here, I will just contribute to docs on how to use glamorous with flow :smiley:

kentcdodds:
Yeah, that sounds awesome. If you want to make contributions to the new site (https://github.com/kentcdodds/glamorous-website) that'd probably be best for the docs. We'll be removing docs from the glamorous repo soonish.

erikras:
I have spent dozens of hours lately working on this for my library, including several conference calls with @thejameskyle. The best way is to have an `index.js.flow` file, next to your published `index.js` file, which does stuff like `declare export var` and `declare export function` for all of the API surface area that you want typed. [Mine looks like this](https://github.com/erikras/redux-form/blob/master/src/index.js.flow). The build step that copies them is [here](https://github.com/erikras/redux-form/blob/master/package.json#L17).

## ⚠️ WARNING ⚠️

You may not import types into your `index.js.flow` from other `.js` files, because they will have all their flow types stripped from them by Babel. If you have common type structures that you want to use in both your exported `index.js.flow` file _and_ in your internally typed code, they will also need to be in `.js.flow` files so that Babel doesn't mangle them. My approach has been to keep my types for each component in a `.js.flow` file with the same name, almost like a `.h` C++ header file, and then they can be imported both into `.js` source files and `.js.flow` type files.

Hope that helps. This is the information that could have saved me _tons_ of time had I known it from the start.

kentcdodds:
Thanks for the input @erikras! I'm good with that. Let's give that a try. Ok @abs-zero?

abs-zero:
yeah sure, I am considering it. Also, I would love to get help from @erikras if I got stuck. :smile:

erikras:
@abs-zero You're welcome to ping me on the Twitters if you get stuck.

erikras:
Oh, and the reason the `index.js.flow` method is superior to `flow-typed` is that the library owner has total control over it. `flow-typed` is just a stop gap measure to deal with libraries that have not yet done it the proper way.

tricoder42:
Just my 2 cents: I also use `index.js.flow` method. It seems to me easier to define types along my code rather than use separate file.

Unfortunately, a lot of people put `/node_modules/` into ignore section of .flowconfig.ini, which makes all type definition distributed with library invisible. It's clearly described in docs, that this isn't recommended and libs with broken types should be ignored explicitly. However, I saw a lot of examples/boilerplates which does that...

valscion:
I've had countless of times when an upgrade of flow broke the shipped types of some library I use. I'd rather not have to upgrade my other dependencies when I upgrade flow.

So even if it'd be easy to ship types with your library, there's no assurance that the types won't break with new flow releases. And if I am required to upgrade other pkgs while I upgrade flow, that makes me a bit sad.

flow-typed seems to be the only solid way of being able to ship updated types for existing versions of any library. This way if v1.2.3 of any pkg had types that worked with flow v0.42 but broke with v0.43, all I'd need to do were to update the types I've got in my own repository and send updates upstream to flow-typed. I could get myself unblocked and not have to forcibly upgrade an old version of a dependency just for the sake of fixing Flow errors.

So my 2 cents is to ship these types in flow-typed, not in the package itself. And this is coming from a heavy user of Flow. Granted, I don't have the perspective of a library maintainer.

abs-zero:
@erikras How do I default export in `index.js.flow`? Also, I don't seem to find any docs on `.js.flow`.

abs-zero:
@kentcdodds @erikras All types I could cover are done but requiring as CJS module isn't working with `index.js.flow`. One has to import as in `const glamorous = require('glamorous').default` although named exports don't work.

Also, where am I supposed to copy flow type files to in a script? Like how in [redux-form](https://github.com/erikras/redux-form/blob/master/package.json#L17), they are copied to `lib` folder.

kentcdodds:
> Also, where am I supposed to copy flow type files to in a script? Like how in redux-form, they are copied to lib folder.

Look at package-scripts.js. We use nps and nps-utils, so you have access to a copy utility with that 👍

tgallacher:
If it's not too late you might also consider `flow-copy-source` plugin which automatically creates the `.flow` files in the output dir for you. 

see: https://github.com/AgentME/flow-copy-source

Also, happy to lend a hand if needed.

kentcdodds:
> requiring as CJS module isn't working with index.js.flow. One has to import as in const glamorous = require('glamorous').default although named exports don't work.

I'm less concerned about this. If someone's using flow, they're probably also using webpack, which will pick up the `module` in `package.json` which points to `dist/glamorous.es.js` :+1:

Someone can make that work later if they want.

kentcdodds:
I just fixed up some git stuff and added a test for the flow typings. Sometimes when I do this, GitHub gets really confused. It's showing this PR includes a TON of totally unrelated commits. But if you pull this branch down yourself you'll see it only has one commit in front of master. I think it should right itself before too long.

In any case, all that's left for this PR now is to figure out how to get the errors in the error snapshot to be correct. Then we can release this as an undocumented beta feature.

I'm still pretty novice at flow though. Anyone with experience wanna try to pull this down and make it great?

abs-zero:
Hey, I am a flow novice, too and I don't use flow any more with my projects these days. So if anyone can help in this, it would be great. 
I think we can make this better with the (flow release 0.53.0)[https://medium.com/flow-type/even-better-support-for-react-in-flow-25b0a3485627]. But I am concerned about the breaking changes.

valscion:
Hi! I thought I'd chime in again about my comment:

> So my 2 cents is to ship these types in flow-typed, not in the package itself. And this is coming from a heavy user of Flow. Granted, I don't have the perspective of a library maintainer.

I just checked https://github.com/facebook/flow/issues/869 again (issue title: Don't typecheck or load files under node_modules/ unless they're imported by flow-typed files). I stand by the idea that shipping types with npm modules risks those types breaking with future flow upgrades, thus requiring the package authors to release a new version and the users of said package to couple updating flow with updating various other packages.

The issue is full of case examples, and the latest one struck to me as quite close to target with this PR: https://github.com/facebook/flow/issues/869#issuecomment-324739727

> Another real-world example:
>
> I updated to the latest version of `flow`, which fixed a `flow` bug. One of my dependencies has a `$FlowFixMe` to workaround the bug. I now get an `Error suppressing comment. Unused suppression` error because the bug has been fixed and the suppression is not necessary with my flow, but the dependency (https://github.com/callemall/material-ui) doesn't get an error with their `flow` version and can't remove that suppression line if they want to support older versions of `flow` that other users may have installed.

---

So that's why I still think it's better to ship types with `flow-typed`. The main reason is that `flow-typed` can enforce the flow version which the typings are compatible with, and can even ship multiple types for the same package, catering to different flow versions.

kentcdodds:
> So that's why I still think it's better to ship types with flow-typed. The main reason is that flow-typed can enforce the flow version which the typings are compatible with, and can even ship multiple types for the same package, catering to different flow versions.

Sounds great to me!

PR #246, opened 2017-07-24
fix: some built-in components will now get the right props as CSS

### LINKED ISSUE #245: bug: anchor tag no longer accepts display prop as css
- `glamorous` version:
- `glamor` version:
- `react` version:

Relevant code.

```javascript
const ui = <A display="flex" />
// renders <a display="flex" />
```

Reproduction:

-- paste your link here --

Problem description:

As noted [here](https://github.com/paypal/glamorous/pull/244#discussion_r129175645):

> At was realized that `A` is a SVG element I guess ([tada](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/a)) which means that all valid SVG props can be forwarded to the `A` tag. This is actually quite unfortunate because we cannot actually tell ahead of time whether an `A` tag is used within an SVG or not. So when you use the `display` prop (as we did in this test), we cannot tell whether you want to forward it on or use it as CSS.
> 
> Today we forward it on rather than using it as CSS. It looks like this changed [11 days ago](https://github.com/jackyho112/react-html-attributes/commit/166ac11d40c7271a414b6c7862ca7f7afa7bf995) in [`react-html-attributes`](https://github.com/jackyho112/react-html-attributes), and it's causing the `display` prop to forward instead of be applied as a css value.

Suggested solution:

We check first if it's an HTML element, if it is then we apply the HTML semantics. If not, then we check if it's an SVG element. Currently the overlap is only: `["a", "script", "style", "svg", "title"]` so I expect that doing this will result in the fewest surprises.

This is a regression so I'm going to push this onto master... Right now....

kentcdodds:
**What**: This adds a check for whether the given `tagName` is an html tag before saying it's svg

**Why**: closes #245

**How**: See the code. Also prettier 😒

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 


src/should-forward-property.js
```diff
@@ -28,7 +29,15 @@ const isCustomAttribute = RegExp.prototype.test.bind(
   new RegExp(`^(data|aria)-[${ATTRIBUTE_NAME_CHAR}]*$`),
 )
 
-const isSvgTag = tagName => supportedSVGTagNames.indexOf(tagName) !== -1
+const isSvgTag = tagName =>
```
kentcdodds: Here's the change that fixes the bug.

src/__tests__/index.js
```diff
@@ -166,37 +166,15 @@ test('style objects can be arrays and glamor will merge those', () => {
   const phoneMediaQuery = '@media (max-width: 640px)'
```
kentcdodds: Reformatting changes only

package.json
```diff
@@ -80,6 +80,7 @@
     "rollup-plugin-uglify": "^2.0.1",
     "tslint": "^5.4.3",
     "tslint-microsoft-contrib": "^5.0.0",
+    "tsutils": "^1.9.1",
```
kentcdodds: We were getting peerDep warnings

PR #250, opened 2017-07-26
feat: add shouldClassNameUpdate for optimizing glamorous components

### LINKED ISSUE #203: add an api for "pure" styles to explicitly be declared.
Problem description:

You want to flag to glamorous to only calculate the className once.

If a dev's styles have **no dynamic values** then why calculate on every render?

Suggested solution:

Add this api

```javascript
glamorous.pure.div({ color: 'blue' })

// or

glamorous.div.pure({ color: 'blue' })

// or

glamorous(
  'div',
  {pure: true}
)
```

I think you can convert normal calls to glamorous to "pure" calls with the existing plugin if this is added.

kentcdodds:
**What**: add `shouldClassNameUpdate` for optimizing glamorous components

**Why**: Closes #203

**How**:
- Add a function to implement this and return `true`/`false` as the case may be
- only compute the className if needed
- Add a test for this functionality

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -23,7 +23,10 @@ function createGlamorous(splitProps) {
   * @param {Object} options helpful info for the GlamorousComponents
   * @return {Function} the glamorousComponentFactory
   */
-  function glamorous(comp, {rootEl, displayName, forwardProps = []} = {}) {
+  function glamorous(
+    comp,
+    {rootEl, displayName, shouldClassNameUpdate, forwardProps = []} = {},
+  ) {
     return glamorousComponentFactory
 
     /**
@@ -42,26 +45,29 @@ function createGlamorous(splitProps) {
        */
       const GlamorousComponent = withTheme(
         (props, context) => {
-          /* eslint no-use-before-define: 0 */
+          const updateClassName = shouldUpdate(props, context)
+
           const {toForward, cssOverrides, cssProp} = splitProps(
             props,
             GlamorousComponent,
           )
 
           // create className to apply
-          const className = getGlamorClassName({
-            styles: GlamorousComponent.styles,
-            props,
-            cssOverrides,
-            cssProp,
-            context,
-            displayName: GlamorousComponent.displayName,
-          })
+          GlamorousComponent.className = updateClassName ?
+            getGlamorClassName({
+              styles: GlamorousComponent.styles,
+              props,
+              cssOverrides,
+              cssProp,
+              context,
+              displayName: GlamorousComponent.displayName,
+            }) :
+            GlamorousComponent.className
 
           return React.createElement(GlamorousComponent.comp, {
             ref: props.innerRef,
             ...toForward,
-            className,
+            className: GlamorousComponent.className,
           })
         },
         {noWarn: true, createElement: false},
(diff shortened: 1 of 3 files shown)


kentcdodds:
I'll go ahead and merge this in a few hours if I don't hear from anyone. Would love a review though!

src/create-glamorous.js
```diff
@@ -81,6 +87,30 @@ function createGlamorous(splitProps) {
         })(GlamorousComponent.styles)
       }
 
+      function shouldUpdate(props, context) {
```
tkh44: This looks like a clever solution, but something about it rubs me the wrong way. It very well may be that I just haven't seen it before. I would suggest calling it something other than state though. I can see this getting confusing when developing and debugging.

kentcdodds: Thanks for the comments!

> This looks like a clever solution, but something about it rubs me the wrong way. It very well may be that I just haven't seen it before.

Do you have suggestions? Happy to make changes.

> I would suggest calling it something other than state though. I can see this getting confusing when developing and debugging.

Good point. Will change it to something else... `meta` maybe?

kentcdodds: Nah, I'll just call it `previous`

tkh44: `previous` works.

kentcdodds: Please take some time to consider how this could be improved. I'm going to merge this into `next` now so we have time to improve it before that's merged into `master` :)

src/create-glamorous.js
```diff
@@ -42,26 +45,29 @@ function createGlamorous(splitProps) {
...
           // create className to apply
-          const className = getGlamorClassName({
-            styles: GlamorousComponent.styles,
-            props,
-            cssOverrides,
-            cssProp,
-            context,
-            displayName: GlamorousComponent.displayName,
-          })
+          GlamorousComponent.className = updateClassName ?
```
DylanMoz: The className and `GlamorousComponent.previous` you attach here is static to the GlamorousComponent, so any instance of the component will share the same className/previous. This seems problematic as it wouldn't support multiple usages of any component. For example: 

```
const MyComponent = glamorous.div(props => { color: props.color })

// Usage
<MyComponent color="green">green</MyComponent>
<MyComponent color="blue">blue</MyComponent>
```

When these two components re-render, the blue component's `shouldClassNameUpdate` function will be given: `prevProps.color` equal `green`, which isn't true, it used to be blue, and is still blue.

kentcdodds: Ah! Nice catch! Hmmm... I'm not certain of the best way to fix it. Would you like to make a PR to try though?

DylanMoz: I believe GlamorousComponent would have to be converted from a functional component to a class... would you be okay with that? We could also potentially _only_ make a stateful component if `shouldClassNameUpdate` is set.

kentcdodds: Actually, it's a little hacky, but we might be able to change [this line](https://github.com/paypal/glamorous/blob/6f6803f15cf2df66a7b035506f5206edff75222b/src/with-theme.js#L70) from this:

```
return ComponentToTheme({...this.props, ...this.state}, this.context)
```

To this:

```
return ComponentToTheme.call(this, {...this.props, ...this.state}, this.context)
```

Then we'd change [the glamorous render function](https://github.com/paypal/glamorous/blob/6f6803f15cf2df66a7b035506f5206edff75222b/src/create-glamorous.js#L47) from an arrow function to a regular function and could reference `this` inside the glamorous function.

A bit hacky, but with comments I think it'd be ok...

kentcdodds: Note that these changes are only in `next` right now, so if you make a PR, make sure you base your branch on `next`.

DylanMoz: That's pretty tricky I didn't think of that. Sounds good.

PR #254, opened 2017-07-27
Optimize getGlamorClassName

tkh44:
By shaping the string class names beginning with 'css-' we can
just give the resulting object to glamor. This offloads almost the
entire workload to glamor from glamorous. Another benefit is that
now there will be at most one `css-[hash]` class name on an element

The only failing test after the modifications was an element that had 2 `css` class names only having one now. The snapshot was updated to reflect this bug fix.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests 
- [x] Ready to be merged 


src/__tests__/__snapshots__/index.js.snap
```diff
@@ -101,13 +101,13 @@ exports[`can use pre-built glamorous components with css prop 1`] = `
...
 exports[`css prop with a className 1`] = `
-.glamor-1,
-[data-glamor-1] {
+.glamor-0,
+[data-glamor-0] {
   color: red;
 }
 
 <section
-  class="glamor-0 glamor-1 abc-123"
```
kentcdodds: :tada: I hadn't even noticed this bug! Might actually help me with something for work too! Woo!

src/get-glamor-classname.js
```diff
@@ -1,4 +1,4 @@
-import {css, styleSheet} from 'glamor'
+import {css} from 'glamor'
```
kentcdodds: I _love_ not having to reach into the glamor to get the `styleSheet`. This is awesome! :clap:

src/get-glamor-classname.js
```diff
@@ -76,21 +86,3 @@ function handleStyles(styles, props, theme, context) {
   }
   return {mappedArgs, nonGlamorClassNames}
 }
-
-function processStringClass(str, mappedArgs, nonGlamorClassNames) {
```
kentcdodds: :clap: for code removal

src/get-glamor-classname.js
```diff
@@ -10,17 +10,23 @@ import {css, styleSheet} from 'glamor'
...
+      glamorlessClassName.push(name)
     }
-    return groups
-  }, {glamorlessClassName: '', glamorStyles: []})
+  })
+
+  return {glamorlessClassName, glamorStyles}
+}
+
+function buildGlamorSrcFromClassName(className) {
```
kentcdodds: Could we add a comment above this line giving a brief explanation of how this black magic works?

src/get-glamor-classname.js
```diff
@@ -33,18 +39,14 @@ function getGlamorClassName({
   theme,
   context,
 }) {
-  const {
-    glamorStyles: parentGlamorStyles,
-    glamorlessClassName,
-  } = extractGlamorStyles(props.className)
   const {mappedArgs, nonGlamorClassNames} = handleStyles(
-    [...styles, parentGlamorStyles, cssOverrides, cssProp],
+    [...styles, props.className, cssOverrides, cssProp],
```
kentcdodds: niiiiiice

kentcdodds: Actually, now that I think of it, this change probably would have worked before as well 😅

kentcdodds:
Oh, and if you _could_ run your fancy benchmarks on this, I'd love to include it in my blogpost if there's any significant perf improvement here...

tkh44:
I didn't see that much of an improvement in rerenders as I thought I would see.

Comparing the 2 fastest runs.

Before
```
- Rerender Duration: 206.05836666666664 ms
- Mount Duration: 180.5808 ms
```

After
```
- Rerender Duration: 196.3391 
- Mount Duration: 166.22899999999998 ms
```

That's about an 8% speedup in mount time. I would be careful to say that this is an 8% speed up though. Benchmarks are stupid 😄 This benchmark does not really test what this change is improving. It uses 1000 random numbers to generate elements with different styles on the page. You rarely hit the cache in that situation. On a real site with 100's of repeat classes and 1000's of elements, I could see this being a real perf win.

kentcdodds:
That's great! I'll just say something like: "And @tkh44 may or may not have made glamorous 50x faster"

src/get-glamor-classname.js
tkh44: Done

PR #255, opened 2017-07-27
feat(withProps): add `withProps` helper API

### LINKED ISSUE #133: Support custom props

kentcdodds:
**What**: This adds the `withProps` helper API

**Why**: Closes #133 

**How**:
- Store the `propsToApply` on the `GlamorousComponent`
- expose a way to create a new glamorous component that combines the `propsToApply` given.
- When rendering, reduce all the `propsToApply` with the `props` given to a single props object

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -31,6 +31,7 @@ function createGlamorous(splitProps) {
       shouldClassNameUpdate,
       forwardProps = [],
       propsAreCssOverrides = comp.propsAreCssOverrides,
+      withProps: basePropsToApply,
     } = {},
   ) {
     return glamorousComponentFactory
@@ -51,6 +52,12 @@ function createGlamorous(splitProps) {
        */
       const GlamorousComponent = withTheme(
         (props, context) => {
+          props = getPropsToApply(
+            GlamorousComponent.propsToApply,
+            {},
+            props,
+            context,
+          )
           const updateClassName = shouldUpdate(props, context)
 
           const {toForward, cssOverrides, cssProp} = splitProps(
@@ -93,6 +100,10 @@ function createGlamorous(splitProps) {
         })(GlamorousComponent.styles)
       }
 
+      function withProps(...propsToApply) {
+        return glamorous(GlamorousComponent, {withProps: propsToApply})()
+      }
+
       function shouldUpdate(props, context) {
         // exiting early so components which do not use this
         // optimization are not penalized by hanging onto
@@ -125,12 +136,14 @@ function createGlamorous(splitProps) {
           rootEl,
           forwardProps,
           displayName,
+          propsToApply: basePropsToApply,
         }),
         {
           withComponent,
           isGlamorousComponent: true,
           previous: null,
           propsAreCssOverrides,
+          withProps,
         },
       )
       return GlamorousComponent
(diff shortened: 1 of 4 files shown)


kentcdodds:
Could I get a review from some of the folks who commented on #133?

@liadbiz, @jfrolich, @marzelin, @wellguimaraes, @ajwhite, @paulmolluzzo, @Pajn, @ericm546, @threepointone

src/create-glamorous.js
```diff
@@ -148,16 +174,54 @@ function createGlamorous(splitProps) {
...
+ * @param {Array} propsToApply an array of propsToApply objects:
+ *   - object
+ *   - array of propsToApply items
+ *   - function that accepts the accumulated props and the context
+ * @param {Object} accumulator an object to apply props onto
+ * @param {Object} props the props that should ultimately take precedence
+ * @param {*} context the context object
+ * @return {Object} the reduced props
+ */
+function getPropsToApply(propsToApply, accumulator, props, context) {
```
luke-john: is this intended to be used elsewhere in the future? otherwise you could drop passing an empty object above and simplify this.

```
function getPropsToApply(propsToApply, props, context) {
    const accumulator = {}
```

kentcdodds: It's used recursively so we need to forward along the accumulator.

kentcdodds:
I would love to merge this today. Hoping to write the blog post to release the lib later this afternoon/evening so we're getting pretty close to the release of v4!

src/create-glamorous.js
```diff
@@ -159,16 +176,54 @@ function createGlamorous(splitProps) {
...
+  // using forEach rather than reduce here because the reduce solution
+  // effectively did the same thing because we manipulate the `accumulator`
```
ajwhite: Yeah, i guess that does make sense. Each iteration in the reducer would just carry over the same object reference. Makes sense.

kentcdodds:
Awesome! Thanks for the review @ajwhite!

PR #257, opened 2017-07-27
feat(typescript): add typing for `withProps` helper APIi

luke-john:
**What and Why**:

This adds the typing required to use the new `withProps` helper API
introduced by https://github.com/paypal/glamorous/pull/255/files.

**How**:
Adding withProps to the GlamorousOptions with a generic type that is also passed to the ComponentFactory allowing the typing to be automatically inferred on usage.

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/other/TYPESCRIPT_USAGE.md
+++ b/other/TYPESCRIPT_USAGE.md
@@ -14,7 +14,7 @@ The typings for
 are complete.
 * using `shouldClassNameUpdate`
 
-```
+```tsx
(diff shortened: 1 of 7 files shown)


luke-john:
Just re reading https://github.com/paypal/glamorous/pull/255/files and it looks like I've missed the `.withProps` function on Glamorous Components.

PR #230, opened 2017-07-13
fix(typescript): complete glamorous component factories

luke-john:
**What and Why**:

This pr follows on from the work in #214  and the discussion on that pr.

BREAKING CHANGE:

These typings use language features of typescript 2.4+.

This fixes a number of issues with the glamorous component factory
typings

* style objects can only be sets of css or svg properties
* style functions can return a string classname or style object
* glamorous arguments can be a string classname
* glamorous arguments can be a StyleArray containing classnames, style objects and style functions.

This also improves the developer experience provided by intellisense                     
with autocomplete once again working.

This closes #212

This also updates the typing for theme to match the new behaviour which landed in #207.

**How**:

Updated the type definitions and the related tests.

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Ready to be merged
- [ ] Added myself to contributors table N/A

These updated typings still have Theme as a seperate parameter to the glamorous component factory function arguments.  I'll try and update this in a seperate pr sometime in the next few days.

I've also taken liberty in the naming of the following argument types as I was unable to find any language describing them in the glamorous documentation.

StyleObject, StyleFunction, StyleFunctionArray, StyleArray

I'd be happy to update these types if someone has suggestions for better names.

PR #262, opened 2017-07-30
chore(typescript): consistant theme handling

luke-john:
Noticed the theme was still being passed as a seperate generic when doing a screen capture for the release notes.

https://github.com/paypal/glamorous/pull/230#issuecomment-318878595

This behaviour is left over from when theme was passed as a seperate argument and had already been removed in some of the typing

**What**:

Removing all typings that had Theme as an additional generic to Props 

**Why**:

Simplifies Theme usage

**How**:

Updated the component factory typings to add overloads for components that make use of Theme

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/src/__tests__/__snapshots__/typescript.js.snap
+++ b/src/__tests__/__snapshots__/typescript.js.snap
@@ -1,120 +1,113 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Typescript expected failures 1`] = `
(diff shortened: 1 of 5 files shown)


kentcdodds:
This has merge conflicts :-(

luke-john:
hmm.. looks like that was due to an extra commit getting caught up in this pr. have fixed

PR #270, opened 2017-08-03
fix: Fix shouldClassNameUpdate to track className per instance, not p…

DylanMoz:
**What**: Fix shouldClassNameUpdate, as it was tracking the previous props from all instances of a particular glamorous component (`previous` and `className` were static).

**Why**: Performance would not be optimized otherwise, unless you were only using one instance of the component.

**How**: Convert the glamorous inner component from an arrow function to a normal function so it can utilize `this`, convert `previous` and `className` to instance variables, and update `with-theme` to pass in its component's `this` as the `this` for the glamorous inner component. In addition, changed the signature of shouldClassNameUpdate from `(props, prevProps, context, prevContext)` to `(props, nextProps, context, nextContext)`, in order to be more inline with the format of other React lifecycle events.

**More Info**: https://github.com/paypal/glamorous/pull/250/files#r130991851

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


DylanMoz:
Awesome!

Can't wait for v4 🙏

PR #273, opened 2017-08-03
Support function as theme prop in ThemeProvider

goodmind:
**What**: Support function as theme prop in ThemeProvider

**Why**: More control over themes merging

**How**: changed `getTheme` in `ThemeProvider`

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged 

It would be better if `glamorous` migrated to `theming` library, but I don't know how to properly handle all this context and `CHANNEL` things


src/theme-provider.js
```diff
@@ -71,7 +84,7 @@ ThemeProvider.contextTypes = {
 }
 
 ThemeProvider.propTypes = {
-  theme: PropTypes.object.isRequired,
+  theme: PropTypes.oneOfType([PropTypes.shape({}), PropTypes.func]).isRequired,
```
ajwhite: What was it about `PropTypes.object` that no longer worked for you?

goodmind: I think `PropTypes.oneOfType([PropTypes.object, PropTypes.func]).isRequired` will be same, I just copied `theming` propTypes

kentcdodds: Yeah, I'd prefer `PropTypes.oneOfType([PropTypes.object, PropTypes.func]).isRequired` :+1: thanks!

src/__tests__/theme-provider.js
```diff
@@ -30,6 +30,57 @@ test('renders a component with theme', () => {
...
+    backgroundColor: bg,
+    color: fg,
+  }))
+  const theme = {
+    fg: 'palevioletred',
+    bg: 'white',
+  }
+  const invertTheme = ({fg, bg}) => ({
+    fg: bg,
+    bg: fg,
```
ajwhite: Idk why, but I love this test you made. Slightly more interesting than the typical unit! 😅

goodmind: This is from `styled-components` examples

src/theme-provider.js
```diff
@@ -16,6 +19,16 @@ class ThemeProvider extends Component {
   // create theme, by merging with outer theme, if present
   getTheme(passedTheme) {
     const theme = passedTheme || this.props.theme
+    if (isFunction(theme)) {
+      const mergedTheme = theme(this.outerTheme)
+      if (!isPlainObject(mergedTheme)) {
+        throw new Error(
+          '[ThemeProvider] Please return an object from your theme function, ' +
```
ajwhite: Kinda like the tag idea here. Do we think it would be useful for the tag mention `glamorous` in here for further visibility into the problem? `[glamorous.ThemeProvider]`, `[glamorous] ThemProvider - ..` etc

src/theme-provider.js
```diff
@@ -16,6 +19,16 @@ class ThemeProvider extends Component {
   // create theme, by merging with outer theme, if present
   getTheme(passedTheme) {
     const theme = passedTheme || this.props.theme
+    if (isFunction(theme)) {
+      const mergedTheme = theme(this.outerTheme)
+      if (!isPlainObject(mergedTheme)) {
+        throw new Error(
```
ajwhite: Do `PropType` validation failures lead to thrown exception in production mode in `React`? I can't recall off the top of my head.

If a `PropType.object` validation would only _warn_, we might want to be more liberal here and do the same. But this needs clarification, as I'm unsure myself

ajwhite:
Just some final thoughts above, nothing pressing. Although I am curious about the prop types validation part

src/theme-provider.js
goodmind: I think `prop-types` validation just don't work in production mode

ajwhite: Yeah @goodmind - that indeed [seems to be the case](https://github.com/facebook/prop-types/blob/0445b8b52edf5e2812f3c6c58922078872b2262e/factoryWithTypeCheckers.js#L166)

ajwhite: Actually - scratch that. That's something else entirely.

kentcdodds:
I think a hard fail on dev and prod is fine here :smile:

PropTypes validation doesn't throw, just console.errors

ajwhite:
My overall thinking was that this is 1:1 to the `PropType.object` validator. Since we can't evaluate the theme until we call it, I'd expect the outcome to be the same as a prop type validation: console error.

I suppose it distills down to: what this case more exceptional than the other?

kentcdodds:
Gotcha. Good point. 🤷‍♂️ I don't think we're worse off having it, so I wasn't concerned about keeping it in.

ajwhite:
Not something I feel passionately about, but would find it consistent 👍

PR #284, opened 2017-08-12
fix(withComponent): maintain component properties

kentcdodds:
**What**: This forwards _all_ properties from the subject component to
the newly created component.

**Why**: To ensure that the new component is basically everything that
the old one was except a new target component to render.

**How**: Spreading component properties.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

I'd really appreciate a review from at least one person on this. The impact is any time you call `.withComponent` it should do what you probably thought it already did. But we _are_ changing behavior here slightly so I'd like to get this reviewed.

My use case was:

```javascript
const FadedText = Text.withProps({fadedExtra: true})
const Description = glamorous(FadedText)({
  margin: '0 0 10px',
}).withComponent('p')
```

The `propsToApply` of `{fadedExtra: true}` was not getting applied to the component returned from `withComponent`. I've verified that this fixes that.

### DIFF
--- a/src/__tests__/create-glamorous.with-component.js
+++ b/src/__tests__/create-glamorous.with-component.js
@@ -38,9 +38,13 @@ test('forwardProps can be overridden for the new component', () => {
 })
 
 test('forwards options', () => {
-  const Text = glamorous.span({color: 'red', fontSize: 20})
+  const propsToApply = {id: 'my-span', faded: true}
+  const Text = glamorous
+    .span({color: 'red', fontSize: 20})
+    .withProps(propsToApply)
   const View = Text.withComponent('div', {displayName: 'View'})
   expect(View.displayName).toBe('View')
+  expect(View.propsToApply).toEqual([propsToApply])
 })
 
 test('resulting component can have its styles extended further', () => {
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -98,10 +98,18 @@ function createGlamorous(splitProps) {
       }
 
       function withComponent(newComp, options = {}) {
-        return glamorous(newComp, {
-          forwardProps: GlamorousComponent.forwardProps,
-          ...options,
-        })(GlamorousComponent.styles)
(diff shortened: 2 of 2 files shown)


src/__tests__/create-glamorous.with-component.js
```diff
@@ -38,9 +38,13 @@ test('forwardProps can be overridden for the new component', () => {
 })
 
 test('forwards options', () => {
-  const Text = glamorous.span({color: 'red', fontSize: 20})
+  const propsToApply = {id: 'my-span', faded: true}
+  const Text = glamorous
+    .span({color: 'red', fontSize: 20})
+    .withProps(propsToApply)
   const View = Text.withComponent('div', {displayName: 'View'})
```
nitin42: Here, extending the old component with the props and styles doesn't generate two classes right ?

src/create-glamorous.js
```diff
@@ -98,10 +98,18 @@ function createGlamorous(splitProps) {
       }
 
       function withComponent(newComp, options = {}) {
-        return glamorous(newComp, {
-          forwardProps: GlamorousComponent.forwardProps,
-          ...options,
-        })(GlamorousComponent.styles)
+        const {forwardProps: fp, ...componentProperties} = GlamorousComponent
+        return glamorous(
```
nitin42: Everything else seems fine to me! This is an effective use case though 👍

src/__tests__/create-glamorous.with-component.js
kentcdodds: Correct. Only one glamor-generated class per glamorous component always.

src/create-glamorous.js
kentcdodds: Thanks!

nitin42:
Seems like glamorous-primitives needs an update (v4 added more features and tweaks in core)!

kentcdodds:
Thanks for the review @nitin42!

PR #290, opened 2017-08-16
feat(withConfig): add (undocumented) `withConfig` for tooling

kentcdodds:
I noticed that
[the styled-components babel plugin](https://github.com/styled-components/babel-plugin-styled-components/blob/37a13e9c21c52148ce6e403100df54c0b1561a88/src/visitors/displayNameAndId.js)
sets the `displayName` by using a `withConfig` function and I thought
that was a pretty good idea. So I'm adding this as an undocumented
feature (because I don't think people will really need to use it outside
the babel plugin)

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "rules": {
       "import/prefer-default-export": 0,
       "react/no-unused-prop-types": 0,
-      "valid-jsdoc": 0
+      "valid-jsdoc": 0,
+      "func-style": 0
     }
   },
   "lint-staged": {
--- a/src/__tests__/.eslintrc
+++ b/src/__tests__/.eslintrc
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "react/prop-types": "off"
+  }
+}
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -23,18 +23,21 @@ function createGlamorous(splitProps) {
   * @param {Object} options helpful info for the GlamorousComponents
   * @return {Function} the glamorousComponentFactory
   */
-  function glamorous(
-    comp,
-    {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -93,7 +93,8 @@
     "rules": {
       "import/prefer-default-export": 0,
       "react/no-unused-prop-types": 0,
-      "valid-jsdoc": 0
+      "valid-jsdoc": 0,
+      "func-style": 0
```
kentcdodds: I got annoyed by this rule. I still prefer function declarations, but I don't care to enforce it anymore.

src/__tests__/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "react/prop-types": "off"
```
kentcdodds: Annoying in tests

src/create-glamorous.js
```diff
@@ -109,7 +112,7 @@ function createGlamorous(splitProps) {
             forwardProps: fp,
             ...options,
           },
-        )(GlamorousComponent.styles)
```
kentcdodds: Turns out we didn't need this at all, and it's actually more performant to not have this because we'd get duplicate styles :scream:

luke-john:
This looks great, looking forward to updating the babel plugin at work. ❤️

As it's an undocumented feature I'll probably leave it untyped for now.

kentcdodds:
Sounds good 👍

soltran:
doesn't look like there are typings for withConfig so there's a type error when I use the babel plugin. I'll try to make a PR (will be my first one!)

Issue #288, opened 2017-08-16
Memory Leak caused by typings


kaoDev:
- `glamorous` version: 4.1.2
- `glamor` version: 2.20.40
- `react` version: 16

Relevant code.

```javascript

```

What you did:
I used glamorous with typescript and the more components I added to the project the slower it got until I reached the `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`

What happened:
the typescript compiler got stuck in a loop

Reproduction:
unfortunately the reproduction is not trivial, but I think any larger project using typescript and glamorous will run into this

-- paste your link here --

Problem description:
The CSSProperties and SVGProperties types both are declared with the helper type 
```typescript
export type SingleOrArray<Properties, T extends keyof Properties> = {
  [P in T]: Properties[P] | Array<Properties[P]>
}
```
from https://github.com/paypal/glamorous/blob/master/typings/helpers.d.ts
this causes some kind of infinite loop eating up huge amounts of memory.

Suggested solution:
For me the fix was to remove this helper, but it was just merged with https://github.com/paypal/glamorous/pull/260 so perhaps @luke-john can say a bit more

luke-john:
Hey @kaoDev, thanks for the bug report and investigation.

We also had oom errors with upgrading to v4 in our component library at work due to a seperate issue with the usage of typescript `extends`.  It seems `extends` are not great with memory at the moment :(.

This `extends` could be replaced by additional interfaces for css and svg properties which does this the long way round, ie.

```ts
export interface CSSPropertiesCompleteArray {
    alignContent?: Array<CSSPropertiesCompleteSingle['alignContent']>
    // ...
}
```

I'm unlikely to get time to open a pr to do this in the coming days, but may get this done over the weekend.  If someone else is able to do this in the meantime (or can think of an alternate solution) that would be great.

kaoDev:
I just realised that there is something else causing the oom problem. When I filed the issue and had a working build I had parts of the v3.25.0 typings in use. Just removing the `SingleOrArray` helper is not enough.

So the real workaround for typescript right now is use v3.25.0

luke-john:
Just had another quick look.

Running the following in the glamorous repo shows a ~80mb difference (less memory usage) when compiling the tests with the component factory overloads working as intended.

`node ./node_modules/.bin/tsc --diagnostics --noEmit`

https://github.com/luke-john/glamorous/pull/2

luke-john:
[...]

PR #291, opened 2017-08-18
fix(typescript): fix component factory typings

luke-john:
**What**:

Glamorous component factory typings are now overloaded to accept
arguments in the following order of matching:

* style arguments with no Props
* style arguments with Props and a Theme
* style arguments with Props

**Why**:

Makes a huge difference to the amount of memory being used by glamorous and hopefully solves the out of memory issue reported at https://github.com/paypal/glamorous/issues/288.

**How**:
Fixing the existing overloads and adding a new one for glamorous components that do not use Props or Theme. 

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/__tests__/__snapshots__/typescript.js.snap
+++ b/src/__tests__/__snapshots__/typescript.js.snap
@@ -7,33 +7,33 @@ exports[`Typescript expected failures 1`] = `
 test/should-fail.test.tsx(16,3): error TS2345: Argument of type '{ fillRule: \\"cat\\"; }' is not assignable to parameter of type 'StyleArgument<SVGProperties, {}>'.
   Type '{ fillRule: \\"cat\\"; }' is not assignable to type '(string | Partial<SVGProperties> | StyleFunction<SVGProperties, {}>)[]'.
     Property 'length' is missing in type '{ fillRule: \\"cat\\"; }'.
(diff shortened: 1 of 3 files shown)


kentcdodds:
Oh! I think this might fix the issues I've been seeing in codesandbox too! Woo! Thanks!

luke-john:
> Oh! I think this might fix the issues I've been seeing in codesandbox too! Woo! Thanks!
 👍 1  

So I just looked into this and I'm not actually sure what's up with the vendored codesandbox monaco editor.

In vscode using just `jsx` you get autocomplete on props and values in style objects, and autocomplete on keys and hinting on values in style functions.

jsx-autocomplete

In codesandbox though the autocomplete doesn't appear to work at all anymore :/.

screen shot 2017-08-21 at 6 49 35 pm

Looks like code sandbox has open sourced the client though ❤️ , so I may be able to investigate further next weekend. https://github.com/CompuIves/codesandbox-client

Ailrun:
@luke-john 
Sorry for bordering you, is there any reason for not to export `StaticStyleArgument` and `StaticStyleArray`, where `StyleArgument` is exported?

luke-john:
Hey @Ailrun , that was an oversight when I created this PR.

In general I've tried to ensure everything is re-exported from the typings entry file.

Also it's not a bother at all, please feel welcome to keep asking questions 👍 .

PR #301, opened 2017-08-23
chore(typescript): improve typescript dx

luke-john:
**What**:

Improvements to the typings to give better parameter hints.

**Why**:

The current typings use types in a number of places which results in enormous parameter hints in vscode and online editors using monaco such as codesandbox and stackblitz.

**How**:

Replaced a number of types with interfaces and removed unneeded Partials.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

Note:

Converting the `GlamorousComponent` from a type -> interface means that when generating definitions users will now need to re-export `GlamorousComponent` rather than `ExtraGlamorousProps`.

## improved-parameter-hints

improved-parameter-hints

## old-parameter-hints

old-parameter-hints

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "tslint": "^5.4.3",
     "tslint-microsoft-contrib": "^5.0.0",
     "tsutils": "^1.9.1",
-    "typescript": "^2.4.2"
+    "typescript": "2.4.2"
   },
   "eslintConfig": {
     "extends": [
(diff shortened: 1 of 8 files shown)


luke-john:
The build error is actually really interesting. It looks like in v2.5+ typescript now provides suggestions for misspellings of props, that's pretty cool 😎.

screen shot 2017-08-23 at 7 11 58 pm

I've rerun the tests with the version of typescript that's on Travis and locked the version of typescript under dev dependencies.

kentcdodds:
There are merge conflicts here now :-(

luke-john:
ahh yeah, should have started off the other branch that was open sorry.

Merge conflicts now resolved

package.json
```diff
@@ -81,7 +81,7 @@
     "tslint": "^5.4.3",
     "tslint-microsoft-contrib": "^5.0.0",
     "tsutils": "^1.9.1",
-    "typescript": "^2.4.2"
+    "typescript": "2.4.2"
```
kentcdodds: Is this intentional?

luke-john: It is yes, typescript 2.5+ introduces suggestions to error messages when it looks like props are misspelt ❤️ .

Typescript doesn't follow semver, so it seemed safer to lock the dev dependency to 2.4.2. anything that work with this _should_ be forwards compatible though :).

kentcdodds:
Thanks!

PR #303, opened 2017-08-24
fix(typescript): make built-in components accept child elements

Kais-DkM:
**What**: Built-in components can now accept children.

**Why**: `React.StatelessComponent` currently requires `children` property to be explicitly stated as its property.
Original typing for glamorous didn't include this, so built-in components couldn't have children.
For example, following code gives type error :
```
<glamorous.Div>
  <glamorous.Span>Hello, world!</glamorous.Span>
</glamorous.Div>
```

**How**: Helper type `Overwrite<T, U>` was added.
With this utility, `children: React.ReactChild` overrides index signature of CSSProperties, and is applied to each built-in components.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


luke-john:
Thanks for fixing this 👍 .

Would you mind adding a couple of tests to `./typings/glamorous.test.tsx` along the following lines so that this isn't broken in future updates to the typings.

```ts
// built-in components accept child elements
(
 <glamorous.Div>
    <glamorous.Span>Hello, world!</glamorous.Span>
</glamorous.Div>
)
```

typings/built-in-glamorous-components.d.ts
```diff
@@ -1,5 +1,11 @@
 import { ExtraGlamorousProps } from './glamorous-component';
 import { CSSProperties } from './css-properties';
+import { Overwrite } from './helpers';
+
+type HTMLComponentProp<T extends HTMLElement> = Overwrite<
+  CSSProperties & ExtraGlamorousProps & React.HTMLProps<T>,
+  { children: React.ReactChild }
```
Ailrun: What React children accepts is React.ReactNode, not React.ReactChild.

Kais-DkM:
After further investigation, I found out that `@types/react` is already inserting `children` property automatically.
However, it's still true that this feature doesn't work in typescript, and it seems that it's somewhat related to typing of `CSSProperties`.
I think it's out of range for this PR, so I'm closing this one.

luke-john:
Thanks for doing the investigation on this and giving it a go.

I _think_ what's needed is something like as follows to allow the CSSProperties index signature to pass for the built in glamorous components children

```ts
// css-properties.d.ts
export interface CSSPropertiesLossyDom {
  [propertyName: string]:
    | string | number | CSSPropertiesComplete | undefined
    | CSSPropertiesLossy
    | React.ReactNode
}

export interface CSSPropertiesDom extends
  CSSPropertiesLossyDom,
  CSSPropertiesComplete,
  CSSPropertiesPseudo
    {}

// /built-in-glamorous-components.d.ts
  A: React.StatelessComponent<CSSPropertiesDom & ExtraGlamorousProps & React.HTMLProps<HTMLAnchorElement>>
```

I'll open up a new issue for this.

Issue #304, opened 2017-08-24
typescript: built-in components don't accept Components as children


luke-john:
Originally reported by @Kais-DkM at https://github.com/paypal/glamorous/pull/303.

```ts
<glamorous.Div>
  <glamorous.Span>Hello, world!</glamorous.Span>
</glamorous.Div>
```

What you did:

Used a Component as a child of a built in glamorous component

What happened:

Typescript errored with `Type '{ children: Element; }' is not assignable to type 'CSSPropertiesDom'.`

Problem description:

The index type for CSSPropertiesLossy is too strict for glamorous component factories

Suggested solution:

Use a custom version of CSSProperties with built in component factories along the following lines

```ts
// css-properties.d.ts
export interface CSSPropertiesLossyDom {
  [propertyName: string]:
    | string | number | CSSPropertiesComplete | undefined
    | Array<CSSPropertiesComplete[keyof CSSPropertiesComplete]>
    | CSSPropertiesLossy
    | React.ReactNode
}

export interface CSSPropertiesDom extends
  CSSPropertiesLossyDom,
  CSSPropertiesComplete,
  CSSPropertiesPseudo
    {}

// /built-in-glamorous-components.d.ts
  A: React.StatelessComponent<CSSPropertiesDom & ExtraGlamorousProps & React.HTMLProps<HTMLAnchorElement>>
```

Other Notes:

For anyone tackling this issue it would be great to add some tests for the built in component factories.

Kais-DkM:
To add more information, currently only built-in components with **single JSX child** failes to pass type checking.
For example, as posted on original post, 
```
<glamorous.Div>
  <glamorous.Span>Hello, world!</glamorous.Span>
</glamorous.Div>
```
fails with error
```
Type '{ children: Element; }' is not assignable to type 'IntrinsicAttributes & CSSProperties & ExtraGlamorousProps & HTMLProps<HTMLDivElement> & { childre...'. Type '{ children: Element; }' is not assignable to type 'CSSProperties'.
Property 'children' is incompatible with index signature.
Type 'Element' is not assignable to type 'string | number | SingleOrArray<CSSPropertiesCompleteSingle, "mask" | "alignContent" | "alignItem...'.
Type 'Element' is not assignable to type '(string | number | {} | (string | number | undefined)[] | ("initial" | "inherit" | "unset" | "fle...'.
Property 'length' is missing in type 'Element'.
```

However, all of the following works without error:
```
<glamorous.Div/>
```
```
[...]

PR #311, opened 2017-08-27
Add ReactChild type to `CSSPropertiesLossy`

Kais-DkM:
**What**: https://github.com/paypal/glamorous/issues/304

**Why**: As described in issue, built-in components couldn't accept single JSX child.

**How**: Original `CSSPropertiesLossy` is renamed as `CSSPropertiesLossyUnrestricted`,
and use `Omit<CSSPropertiesLossyUnrestricted, 'children'>` as `CSSPropertiesLossy`.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 


src/__tests__/__snapshots__/typescript.js.snap
```diff
@@ -1,51 +1,41 @@
...
+test/should-fail.test.tsx(10,3): error TS2345: Argument of type '{ fillRule: \\"cat\\"; }' is not assignable to parameter of type 'StyleArgument<SVGProperties, {}>'.
   Type '{ fillRule: \\"cat\\"; }' is not assignable to type '(string | SVGProperties | StyleFunction<SVGProperties, {}>)[]'.
     Property 'length' is missing in type '{ fillRule: \\"cat\\"; }'.
 test/should-fail.test.tsx(16,3): error TS2345: Argument of type '{ fillRule: \\"cat\\"; }' is not assignable to parameter of type 'StyleArgument<SVGProperties, {}>'.
   Type '{ fillRule: \\"cat\\"; }' is not assignable to type '(string | SVGProperties | StyleFunction<SVGProperties, {}>)[]'.
     Property 'length' is missing in type '{ fillRule: \\"cat\\"; }'.
-test/should-fail.test.tsx(22,3): error TS2345: Argument of type '() => { fillRule: string; }' is not assignable to parameter of type 'StyleArgument<SVGProperties, {}>'.
-  Type '() => { fillRule: string; }' is not assignable to type '(string | SVGProperties | StyleFunction<SVGProperties, {}>)[]'.
-    Property 'push' is missing in type '() => { fillRule: string; }'.
```
luke-john: There are a lot of missing test failures in this file.

Looks like something has introduced a regression in the type safety.

typings/svg-properties.d.ts
```diff
@@ -282,13 +282,15 @@ type SVGPropertiesComplete = SingleOrArray<
...
-export interface SVGPropertiesLossy {
+export interface SVGPropertiesLossyUnrestricted {
   [propertyName: string]:
     | string | number | SVGProperties | undefined
     | Array<SVGPropertiesCompleteSingle[keyof SVGPropertiesCompleteSingle]>
-    | SVGPropertiesLossy
+    | Omit<SVGPropertiesLossyUnrestricted, 'children'>
 }
 
+type SVGPropertiesLossy = Omit<SVGPropertiesLossyUnrestricted, 'children'>;
```
luke-john: The underlying issue here is that this indexer is not matching `{ children?: React.ReactNode }`.

So rather than omit from this indexer, we should be able to add `React.ReactNode` to the property union in the original `SVGPropertiesLossy`.

This strategy should also fix the tests which are no longer failing.

src/__tests__/__snapshots__/typescript.js.snap
```diff
@@ -1,51 +1,41 @@
...
+  Object literal may only specify known properties, and '':hover'' does not exist in type 'StyleArgument<SVGProperties, {}>'.
+test/glamorous.test.tsx(52,3): error TS2345: Argument of type '{ overflowWrap: string; display: (\\"block\\" | \\"flex\\")[]; }' is not assignable to parameter of type 'StyleArgument<CSSProperties, {}>'.
+  Object literal may only specify known properties, and 'overflowWrap' does not exist in type 'StyleArgument<CSSProperties, {}>'.
+test/glamorous.test.tsx(57,3): error TS2345: Argument of type '{ textAlign: string; display: (\\"marker\\" | \\"block\\")[]; }' is not assignable to parameter of type 'StyleArgument<SVGProperties, {}>'.
+  Object literal may only specify known properties, and 'textAlign' does not exist in type 'StyleArgument<SVGProperties, {}>'.
+test/glamorous.test.tsx(66,3): error TS2345: Argument of type '{ ':active': { \\"fontSize\\": number; }; '&:active': { \\"fontSize\\": number; \\"textAlign\\": string; }; }' is not assignable to parameter of type 'StyleArgument<CSSProperties, {}>'.
```
luke-john: These failures should not be occurring (the usage they are testing is valid).

Kais-DkM:
@luke-john 
Oh, I'm sorry that I haven't read testing log carefully. I'll fix this asap.

Also, the reason I chose `Omit` rather than adding `React.ReactNode` to union is because I wanted to block assigning `ReactNode` typed values to property.
For example, If I put `ReactNode` to union, something like
`<glamorous.Div color={<div />}>` would be possible.

luke-john:
That shouldn't actually be a problem as `color` will be receiving a type of string from CSSPropertiesComplete so you'd get a type failure still.

Kais-DkM:
@luke-john 
I think your explanation is right.
How about this one? : `<glamorous.Div borderRadius={<div />}>`
Since borderRadius is not a key of `CSSPropertiesCompleteSingle`, I think this would pass type checking.

luke-john:
Ahh yes, that would pass type checking.

Actually currently the following will pass as well due to that indexer

* `<glamorous.Div borderRadius={'also incorrect'}>`
* `<glamorous.Div borderRadius={1}>`

Actually I suspect using `CSSPropertiesCompleteSingle` in `built-in-glamorous-components` will solve this problem without having to make changes to `CSSProperties`.

Kais-DkM:
@luke-john 
I think it's not a good solution because it would block applying properties that are not included in `CSSPropertiesCompleteSingle` (like `borderRadius`).

Also, I think two examples you gave is valid cases.
Since we can't get type information about those kind of properties by its name(again, like `borderRadius`), we shouldn't restrict its type into something.

luke-john:
Ahh sorry, I misunderstood what you were getting at there, I thought borderRadius was misspelt rather than missing from `CSSPropertiesCompleteSingle`.

The fact `borderRadius` is missing from `CSSPropertiesCompleteSingle` should probably be considered as a bug (it looks like this is a bug we've inherited from when we vendored the react typings for `CSSProperties`).

Using `CSSProperties` rather than `CSSPropertiesCompleteSingle` will mean that misspellings for css property names are not picked up.

*  `<div textRendering="optimizeLegibility" />`
*  `<div textRenderng="optimizeLegibility" />`

Regardless of which approach we take here, I think it would be great to do an audit of `CSSProperties` and see if there's some way to at least partially automatically generate the definitions from the spec or somewhere else :).

Kais-DkM:
@luke-john 
After some investigation, I concluded that there are no viable solution for now other than adding type to `CSSPropertiesLossy` union.
Added type was `React.ReactChild` rather than `React.ReactNode` because `{}`, which is included in `React.ReactNode` causes some failure tests to be passed.
For our case, the problem was component with only one child, so it wouldn't have any further problem.

It would be really nice if someone fix this with better solution later....

luke-john:
This looks good to land to me, thanks again for all your work on this 💗.

PR #320, opened 2017-09-03
[WIP] feat(typescript): use JSX.IntrinsicElements for properties

luke-john:
This replaces the use of `React.HTMLProps` and `React.SVGAttributes` with `JSX.IntrinsicElements`.

Inspired by https://github.com/styled-components/styled-components/pull/1084

**What**:

Replaces the use of `React.HTMLProps` and `React.SVGAttributes` with `JSX.IntrinsicElements`.

**Why**:

The use of these came from when the typings were first based off `styled-component`.

**How**:

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests
- [ ] Ready to be merged

I haven't tested this thoroughly, so whilst it all seems fine, I'm not sure this is ok to land just yet.  I'll try and do some testing with a large glamorous codebase in the coming days.

PR #321, opened 2017-09-03
feat(preact): add native support for preact

### LINKED ISSUE #85: Make glamorous work out of the box with preact
Original issue: https://github.com/developit/preact/pull/653#issuecomment-297111106

At the moment there are a couple of snags while working with preact:

1. preact doesn't handle the `autoFocus` attribute in the same way that react does, and, at the moment, [does not declare that `autoFocus` (it uses `autofocus`)](https://github.com/developit/preact/pull/653) is a valid prop on html intrinsic elements in its typescript definition. Glamorous, on the other hand, only whitelists `autoFocus` as a prop to be passed on to the intrinsic element, so it does not pass `autofocus` on. @kentcdodds suggested using `forwardProps`, but I'm not sure how to use that in this case just yet and it seems like it'd be a hack.
2. React is a peer-dependency of glamorous, which means you get a warning if you don't have it installed.
3. Glamorous requires `preact-compat`. Ideally it would work w/o it.

marzelin:
**What**: Adds out-of-the-box support for `preact`

**Why**: Even though `glamorous` can be used in `preact` project, there are a few issues:
- there are some component props that are allowed in `preact` but invalid in `react`.
- React is a peer-dependency of glamorous, which means you get a warning if you don't have it installed.
-  `preact-compat` is required to be able to use `glamorous`

**How**:
Build process creates **preact** directory that stores files dedicated for this library. To use them one needs to import glamorous from **glamorous/preact**.

`preact` files differ from normal files in a following way:
- preact-specific props are added to props list in **./react-props**
- `PropTypes` are neutralized (`preact` doesn't use them) and `Children` polyfilled in **./react-compat**.
- All `react` imports are replaced with `preact` (at build time).

Other changes:
`react` is removed from peerDependencies in **package.json**
A shim of **package.json** is added to `preact` dir (at build time) to allow for importing files from there.
`typings` folder is copied to **preact** folder to provide type definitions for `typescript`

closes #85 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged 


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
kentcdodds: Could you add this file to the gitignore? I don't like editor-specific settings in my repositories 😀

other/preact.package.json
```diff
@@ -0,0 +1,19 @@
+{
```
kentcdodds: What's this for?

luke-john:
> typings folder is copied to preact folder to provide type definitions for typescript

I'm not familiar with preact or it's typings, but have you tested that this works, seems like there's a reasonable chance you'll get type failures with the react typings conflicting with preact ones.

If this is a problem I _think_ https://github.com/paypal/glamorous/pull/320 should improve the situation.

I'm unlikely to be able to spend much time looking into how the typings play with preact this week, but love the idea that the typings can be shared ❤️.  Is there something like create-react-app for preact I can give this a quick spin?

kentcdodds:
I've published this as the next beta!

```
~/Desktop/glamorous (master)
👻  $ npm publish --tag preact
+ glamorous@4.8.0-beta.0
```

Please try it out and let me know whether it works.

marzelin:
@luke-john there's [preact-cli](https://github.com/developit/preact-cli) but it contains a lot of packages, including `preact-compat`. I've created [minimal typescript+preact+glamorous project](https://github.com/marzelin/glamorous-beta-tests) and you're right, typescript throws errors since react is also imported in type definitions, unfortunately.

Apart from that everything seems to work. [Here's a playground](https://codepen.io/marzelin/pen/gxqJxE) that uses beta umd build.

other/preact.package.json
marzelin: `package.json` file needs to be present in `preact` folder since node.js looks into it to figure out which file to import. This file is used as a blueprint for it. Perhaps I could use main package.json file for this as well :thinking:

.vscode/settings.json
marzelin: Sure. I need this file because my prettier rules are different and they're applied on each save. I'll keep it only in my fork.

luke-john:
Thanks for that @marzelin.

I _think_ a nice way to solve this could be to have a file which re exports aliases of the react interfaces used by the typings. And then as part of the copy step for preact replace that file with one that has the preact equivalents.

src/__tests__/preact.js
```diff
@@ -0,0 +1,48 @@
+beforeEach(() => {
+  process.env.PREACT = true
+  jest.resetModules()
+  jest.mock('react', () => ({}))
```
kentcdodds: Why is this necessary? Should it be: `jest.mock('react', () => require('preact'))` instead?

marzelin: Yeah, I've added `preact` as a dev dependency after all. Good idea :clap:

package-scripts.js
```diff
@@ -44,29 +47,67 @@ module.exports = {
...
+          'build.es.preact',
+          'build.cjs.preact',
+          'build.umd.main.preact',
+          'build.umd.min.preact',
+          'build.es.preact.tiny',
+          'build.cjs.preact.tiny',
+          'build.umd.main.preact.tiny',
+          'build.umd.min.preact.tiny'
+        ),
+        copy("'package.json' 'preact'")
```
kentcdodds: Hmmm... What does this do exactly?

I'm a little confused about the build process. When everything's finished, what are we left with?

marzelin: It creates a directory `preact` in which the are all builds for preact as well as a copy of `package.json` from the root directory. No typings at this time since current type definitions are react-dependent.

kentcdodds: I see! So, someone using the package would do:

```
import glamorous from 'glamorous/preact'
```

Is that right? That makes good sense. For `downshift` I push everything into the [`dist/`](https://unpkg.com/downshift@1.4.0/dist/) and have `.preact` filename, but I think I like this better. Thanks!

kentcdodds:
I'm working on this right now. I want to add kcd-utils to this project, so I'm just going to stick that in as part of this PR :)

kentcdodds:
I literally spent all day on this :joy:

But I think I've finally got it working a great way.

Thank you for all your work on it @marzelin!

kentcdodds:
Ok! I'll merge this and the docs PR as soon as I get one person to confirm that this works with preact-cli:

```
~/Desktop/glamorous (master)
🎉  $ npm publish --tag beta
+ glamorous@4.9.0-beta.0
```

kentcdodds:
Not quite there. I just tested it and something's messed up. Should be a quick fix...

kentcdodds:
Awesome! I've verified that `glamorous@4.9.0-beta.1` (this PR) works great with `preact-cli`. `preact-cli` will pick up the `preact/dist/glamorous.esm.js` file when you do:

```js
import glamorous from 'glamorous/preact'
```

:tada:

kentcdodds:
[Tweeted](https://twitter.com/kentcdodds/status/907749455331561479). Sorry, I couldn't find your twitter handle :-/

Thanks so much!!

kentcdodds:
Thanks so much for your help! I've added you as a collaborator on the project. Please make sure that you review the CONTRIBUTING.md and other/MAINTAINING.md files (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

marzelin:
Awesome :tada:. Thanks @kentcdodds

PR #333, opened 2017-09-20
fix(propTypes): allow object as className

### LINKED ISSUE #332: Glamor css prop w/ glamorous fails PropType validation
Hi there!

- `glamorous` version: 4.9.2
- `glamor` version: 2.20.40
- `react` version: 15.6.1

Relevant code.

```jsx
import glamorous from 'glamorous';

const GlamDiv = glamorous.div({
  color: 'orange'
});

class App extends Component {
  render() {
    return (
      <div>
        <div css={{ textDecoration: 'underline' }}>glamor css prop</div>
        <GlamDiv>glamorous div</GlamDiv>
        <GlamDiv css={{ textDecoration: 'underline' }}>
          glamor css prop on glamorous div. Look at console.
        </GlamDiv>
      </div>
    );
  }
}
```

What you did:

I've been using glamor + glamorous and I recently discovered glamor's `css` prop (https://github.com/threepointone/glamor/blob/master/docs/createElement.md). It's awesome! However, I've found that it doesn't play well with `glamorous`. 

What happened:

screenshot

Reproduction:

Here's a reproduction in codesandbox: https://codesandbox.io/s/mqqq7k5z18

I also made this repo before I realized I _could_ use codesandbox: https://github.com/Tibfib/glamor-glamorous-css-bug

Problem description:

When using both glamorous and glamor's `css` prop there is a PropTypes validation error. For all I can tell it works perfectly–it just complains about PropTypes.

I do think there's a use case for using both glamorous and the css prop.

Suggested solution:

Play nicely with `glamor`'s CSS prop. I think this is a `glamorous` issue but it could be an issue with how `glamor` does the `css` prop. 

Thanks for reading!

kentcdodds:
**What**: This adds `object` as an option for `className` to allow for passing `glamor.css` to the `className` prop.

**Why**: This closes #332

**How**: `PropTypes.oneOfType`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

cc @Tibfib and @marzelin

### DIFF
--- a/src/__tests__/create-glamorous.with-component.js
+++ b/src/__tests__/create-glamorous.with-component.js
@@ -60,6 +60,6 @@ test('resulting component can have its styles extended further', () => {
     'overriding styles via css prop',
   )
   expect(
-    render(<View className={glamor.css({color: 'green'}).toString()} />),
+    render(<View className={glamor.css({color: 'green'})} />),
   ).toMatchSnapshot('overriding styles via className')
 })
--- a/src/create-glamorous.js
+++ b/src/create-glamorous.js
@@ -95,7 +95,9 @@ function createGlamorous(splitProps) {
       )
 
       GlamorousComponent.propTypes = {
-        className: PropTypes.string,
+        // className accepts an object due to glamor's css function
+        // returning an object with a toString method that gives the className
+        className: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
         cssOverrides: PropTypes.object,
         innerRef: PropTypes.func,
         glam: PropTypes.object,


marzelin:
lgtm, but allowing for assigning objects to `className` doesn't feel right.

kentcdodds:
I agree it doesn't feel right, and normally you shouldn't, but considering we claim to have great integration with glamor we should at least help people avoid this confusion. I've seen it over and over again and I'm tired of fielding questions here. If someone mistakenly passes an object, the error will be more obvious...

PR #336, opened 2017-09-24
fix(typescript): remove deprecated npm script

### LINKED ISSUE #335: Fix test:ts
See [this build](https://travis-ci.org/paypal/glamorous/builds/279108767) and you'll see the `test:ts` output has a whole bunch of messages that I think mean something's wrong. I'm not sure what's up though.

luke-john:
The `test:ts` script was being used previously to test that the
typescript definitions were working correctly.  These were moved
into the main tests to allow snapshots to be maintained with
extra tests which ensure the defintions fail on incorrect usage
of glamorous.

**What/How**:

remove deprecated npm script `test:ts`

**Why**:

It now fails with the should-fail test file.

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged

closes #335

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test:update": "npm run test:cover -s -- --updateSnapshot",
     "test:build": "kcd-scripts test --config other/jest.config.js --no-watch",
     "build-and-test": "npm run build -s && npm run test:build -s",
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks!

PR #340, opened 2017-09-28
Include filter props

johnjesse:
**What**:

Added filterProps to the glamorousOptions interface and the propsAreCSSOveridesOptions interface.
Also changed the forwardProps on both interfaces so it is typed to `string[]`

**Why**:

They are methods missing on the interfaces, and are causing compiler errors when running with typescript

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

I added the pre-commit line in the opt-in file, but didn't actually get told why my commits failed so I ended up commiting under the -no-verify flag.


johnjesse:
Failed because the snapshot was updated? But the contributing log said to run 'npm run test:update' and commit any snapshot changes...

Also, I don't really understand why the snapshot changed when I only added types to some interfaces

kentcdodds:
Whoops! Wrong button!

kentcdodds:
The snapshot changes look totally unrelated to your changes. I'm not sure why you needed to update the snapshots locally 🤔

johnjesse:
I'll back them out, but here https://github.com/paypal/glamorous/blob/master/CONTRIBUTING.md it does say to run the tests locally, I guess I misunderstood the direction and you only need to run with the update flag if you are meant to be updating snapshots?

kentcdodds:
No, I think you didn't do anything wrong. There's something messed up with glamorous's tests right now. I'm working on it.

kentcdodds:
I'm afraid that I've run out of time to work on this. I'll try to get back to it eventually, but I can't make any promises (we just had baby # 4 and my kids are wondering why daddy isn't playing with them while Mom takes care of the baby).

If anyone wants to step in here to figure out what's going on that'd be great.

luke-john:
Just had a look into this, the snapshot mismatchs look like they are related to the version of node you run the tests with.

If you run them with node 6 you get the output as seen in this PR, if you run them with node 8 you get the output as seen on the current master branch.

@johnjesse would you be able to rerun the tests with node 8 to redo the snapshots?

johnjesse:
@luke-john, thanks for looking into it. I'll re-run the tests with node 8 as soon as I get the chance

johnjesse:
Hey, so nothing has happened on this for a while, I know the checks haven't reported, do you want me to do anything else, squash the commits etc...

PR #345, opened 2017-10-18
Update README.md

shreyas-a:
Center aligned badges


kentcdodds:
Thanks for this, but I actually don't like the way it looks with the badges centered.

Thanks anyway! I appreciate your willingness to help!

PR #351, opened 2017-10-26
Allow to forward innerRef to underlying component via `forwardProps`

### LINKED ISSUE #349: Suggestion: allow `innerRef`-prop to be passed to underlying component via option-flag
I'm working on a project with a big `Ui` File which exports a lot of components like Buttons, Boxes, Text-Components, etc.

I'd like those to behave as similarily as possible. I.e. I'd like them all to use `innerRef` to be able to access the underlying _DOM_-Node.

For wrapping plain html-tags (`glamorous('div')({...})`) this works as expected. 
Now if I want to wrap a Component like a React-Router `Link` I can do `const WrappedLink = glamorous(Link)({})`

If I write `<WrappedLink innerRef={...}/>` `innerRef` gives me the reference to the `Link` Component. However I would like to do something like this:

```
const WrappedLink = glamorous(Link, {forwardProps: ["innerRef"]})({})
```

such that `innerRef` now is passed to the `Link` component yielding its underlying DOM-Node.

I believe that fixing [this line](https://github.com/paypal/glamorous/blob/f8d833619547e64506fabb40dead6e6992a3347a/src/create-glamorous.js#L89) could resolve this issue.

Do you believe it's worthwhile pursuing? If so I'm happy to do a PR :)

danielberndt:
**What**:

Fixes #349 

This PR allows users to write e.g. 
```const WrappedLink = glamorous(Link, {forwardProps: ["innerRef"]})({...})```
This will forward `innerRef` to the underlying component.

`<WrappedLink innerRef={}/>` will now yield whatever `Link`'s `innerRef` returns.

**Why**:

This allows you write more consistent components. You now can more easily guarantee that `innerRef` within your code will always return DOM Nodes, no matter whether they got wrapped by glamorous or now

**How**:

Both, the `split-props.js` as well as the `tiny.js` now verify if `innerRef` should be forwarded. If this is the case, glamorous won't attach it's own `innerRef` handler.

**Checklist**:

- [ ] Documentation N/A (not sure whether this applicable to the PR, since the Documentation seems to live somewhere else?)
- [x] Tests
- [x] Ready to be merged

fyi `src/__tests__/theme-provider.js` tests are failing when checking out the project and installing everything from scratch. I'll create an issue.

### CODE REVIEW

**Comment by marzelin** [comment]
Nice. Can you make it work for the following test cases as well:
```js
test("should pass inner ref to underlying component if forwarded and rootEl is set", () => {
  let nodeType = null;
  const getRef = n => (nodeType = n);

  class InnerComp extends React.Component {
    render() {
      return (
        <div>
          <span ref={this.props.innerRef} />
        </div>
      );
    }
  }

  const Comp = glamorous(InnerComp, {
    rootEl: "span",
    forwardProps: ["innerRef"]
  })({
    marginLeft: "24px"
  });

  mount(<Comp innerRef={getRef} blue />);

  expect(nodeType).toBeInstanceOf(HTMLElement);
  expect(nodeType.tagName).toBe("SPAN");
});

test("should pass inner ref to underlying component if forwarded and filterProps are specified", () => {
  let nodeType = null;
  const getRef = n => (nodeType = n);

  class InnerComp extends React.Component {
    render() {
      return (
        <div>
          <span ref={this.props.innerRef} />
        </div>
      );
    }
  }

  const Comp = glamorous(InnerComp, {
    filterProps: ["noname"],
    forwardProps: ["innerRef"]
  })({
    marginLeft: "24px"
  });

  mount(<Comp innerRef={getRef} blue />);

  expect(nodeType).toBeInstanceOf(HTMLElement);
  expect(nodeType.tagName).toBe("SPAN");
});
```

danielberndt:
Good point @marzelin! I wasn't quite sure whether the other code paths were relevant as well, now I changed it in a way that it gets executed in any case.

src/split-props.js
```diff
@@ -13,6 +13,10 @@ export default function splitProps(
   },
   {propsAreCssOverrides, rootEl, filterProps, forwardProps},
 ) {
+  // forward innerRef if user wishes to do so
+  if (innerRef !== undefined && forwardProps.indexOf('innerRef') !== -1) {
```
marzelin: what's the point of checking `innerRef`'s value here?

src/tiny.js
```diff
@@ -1,16 +1,23 @@
...
-  // component ever
-  ...rest
-}) {
+function splitProps(
+  {
+    css: cssProp,
+    // these are plucked off
+    theme, // because they
+    className, // should never
+    innerRef, // be forwarded
```
marzelin: comments here are not relevant to `innerRef` any more. I'd move the prop above the comments.

danielberndt: 👍

src/split-props.js
danielberndt: Performance. There's no need to loop through the array if `innerRef` is empty anyway.
It's in glamorous' "hot path", but I'm not sure it's really that critical. Happy to leave it out.

danielberndt:
Okay, the `__tests__/index.js` is too long now. Should I move all ref-related tests into their own file?

src/split-props.js
marzelin: I conveniently leave decisions concerning optimizations vs longer code to @kentcdodds :)

kentcdodds:
@danielberndt, ummmm... Yeah, could you do that actually, that'd be nice. Could call it `__tests__/refs.js` 🤷‍♂️ 

Thanks!

src/split-props.js
kentcdodds: I'm fine with it as-is :+1:

Thanks!

src/__tests__/index.js
```diff
@@ -372,6 +372,107 @@ test('should receive inner ref if specified', () => {
   expect(getRef).toHaveBeenCalled()
 })
 
+test('should pass inner ref to underlying component if forwarded', () => {
+  let nodeType = null
```
kentcdodds: Instead of `nodeType`, let's call this `node` :+1:
Same with other similar tests.

src/__tests__/index.js
```diff
@@ -372,6 +372,107 @@ test('should receive inner ref if specified', () => {
...
+      return (
+        <div>
+          <span ref={this.props.innerRef} />
+        </div>
+      )
+    }
+  }
+
+  const Comp = glamorous(InnerComp, {
+    rootEl: 'span',
```
kentcdodds: It doesn't technically matter, but `rootEl` is actually a `div` in this case. Should probably fix it just to avoid confusing future contributors.

marzelin: well, it kind of depends where we pass props received from glamorous, isn't it?  In this case we pass them neither on `div` nor `span`. :thinking:

danielberndt: fixed 👍

src/__tests__/index.js
danielberndt: done!

kentcdodds:
Thanks so much for your help! I've added you as a collaborator on the project. Please make sure that you review the `other/MAINTAINING.md` and `CONTRIBUTING.md` files (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

danielberndt:
Thanks for your feedback and thanks for your trust @kentcdodds.
I really do appreciate your way of managing Open Source projects and do hope that many more projects will adapt this style of leadership!

PR #363, opened 2017-12-21
Fixed nested themes not being republished on outer theme changes

Andarist:
**What**:

This fixes a bug when having nested `ThemeProvider`s and updating outer theme the change is not reaching the `StyledComponent`
```jsx
<ThemeProvider theme={{}}>
	<ThemeProvider theme={{}}>
		<StyledComponent>
	</ThemeProvider>
</ThemeProvider>
```

**Why**:

This is a bug fix.

**How**:
Using publish in the nested `ThemeProvider`s subscriptions.

**Checklist**:
- [x] Tests
- [x] Code complete

Repro provided by @mitchellhamilton - https://codesandbox.io/s/6j42qx5z23

### DIFF
--- a/src/theme-provider.js
+++ b/src/theme-provider.js
@@ -40,6 +40,11 @@ class ThemeProvider extends React.Component {
 
   setOuterTheme = theme => {
     this.outerTheme = theme
+    this.publishTheme()
+  }
+
+  publishTheme(theme) {
+    this.broadcast.setState(this.getTheme(theme))
   }
 
   componentDidMount() {
@@ -53,13 +58,12 @@ class ThemeProvider extends React.Component {
     // set broadcast state by merging outer theme with own
     if (this.context[CHANNEL]) {
       this.setOuterTheme(this.context[CHANNEL].getState())
-      this.broadcast.setState(this.getTheme())
     }
   }
 
   componentWillReceiveProps(nextProps) {
     if (this.props.theme !== nextProps.theme) {
-      this.broadcast.setState(this.getTheme(nextProps.theme))
+      this.publishTheme(nextProps.theme)
     }
   }
 
--- a/src/__tests__/__snapshots__/theme-provider.js.snap
+++ b/src/__tests__/__snapshots__/theme-provider.js.snap
@@ -46,6 +46,32 @@ exports[`merges nested themes 1`] = `
 </div>
 `;
 
+exports[`propagates theme updates through nested ThemeProviders 1`] = `
+.glamor-0,
+[data-glamor-0] {
+  background-color: black;
+  color: red;
+}
+
+<ThemeProvider
+  theme={
+    Object {
+      "bg": "black",
+    }
+  }
+>
+  <ThemeProvider
+    theme={[Function]}
+  >
+    <glamorous(div)>
+      <div
+        className="glamor-0"
+      />
+    </glamorous(div)>
+  </ThemeProvider>
+</ThemeProvider>
+`;
+
 exports[`renders a component with theme 1`] = `
 .glamor-0,
 [data-glamor-0] {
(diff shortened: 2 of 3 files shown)


src/theme-provider.js
```diff
@@ -53,13 +58,12 @@ class ThemeProvider extends React.Component {
     // set broadcast state by merging outer theme with own
     if (this.context[CHANNEL]) {
       this.setOuterTheme(this.context[CHANNEL].getState())
-      this.broadcast.setState(this.getTheme())
```
kentcdodds: Can you explain why this was removed? It seems unrelated.

Andarist: publishTheme method is doing a setState and setOuterTheme uses now publishTheme under the hood

Andarist:
@kentcdodds do u have any other reviewer in mind? The chane is pretty straightforward and the codebase is tested so it shouldnt break anything ;)

src/theme-provider.js
```diff
@@ -53,13 +58,12 @@ class ThemeProvider extends React.Component {
...
     }
   }
 
   componentWillReceiveProps(nextProps) {
     if (this.props.theme !== nextProps.theme) {
-      this.broadcast.setState(this.getTheme(nextProps.theme))
+      this.publishTheme(nextProps.theme)
```
paulmolluzzo: Pardon my ignorance, but how has this made an effectual change? `publishTheme` is calling `broadcast.setState(this.getTheme(nextProps.theme))` which is was was here before. It's only been moved to a separate method.

src/theme-provider.js
paulmolluzzo: I think this is the real fix, and `publishTheme` is only incidentally changed.

This line was setting the state with no argument, which defaults to the `props.theme` here https://github.com/paypal/glamorous/blob/cb6de3abe6b3a7e2643828f4239e0487d4f80455/src/theme-provider.js#L21, meaning it's not setting the theme to the *next* theme (it's just resetting it to the theme it already has).

A lot of the theming work was done by @vesparny, so if I'm wrong hopefully he can correct me. 😄

src/theme-provider.js
Andarist: oh, this part is actually not a fix of any sorts, i just have extracted it to separate method so it would be easier to reuse it in `setOuterTheme`

paulmolluzzo: OK, thanks for clarifying.

src/theme-provider.js
Andarist: Yeah, this part of defaulting to `this.props.theme` is a making a control flow a little bit iffy but I left this part untouched. Personally I would prefer explicit arguments, so that `publishTheme` would always have to receive `getTheme`'s result as argument which in turn would always be called with `this.props.theme` if necessary.

Also I've noticed a different small issue too - broadcast is initialized as:
```js
broadcast = brcast(this.props.theme)
```
but that gives it untrue initialState in case of nested ThemeProvider. I think creation of the `broadcast` should be delayed.

paulmolluzzo: > but that gives it untrue initialState in case of nested ThemeProvider. I think creation of the broadcast should be delayed.

I'm pretty sure I follow, but to be clear, are you suggesting moving it to something like `componentWillMount`? With this PR, the theme that was (possibly) merged with the outer theme is now going to be set properly inside `componentWillMount()`, so that untrue initial state won't really ever be in effect.

Did I follow your point correctly? If not any clarification would be great. 🙌

Andarist: Yeah, that was exactly my point :)

paulmolluzzo: Cool, thanks! It makes sense to me. Maybe you want to do that as a follow-up PR?

Andarist: Sure thing, would prefer to create it once we settle on this one to avoid potential conflicts

kentcdodds: Thanks for the discussion here!

PR #369, opened 2018-01-04
fix(typings): allow to have primitives as children for built-in

denis-sokolov:
**What**: Incorrect typing for built-in components.

**Why**: Because today in client code I have to add a manual typing override: `<Div>{foo as any}</Div>` if `foo` can be `null`.

**How**: By using the correct type from React library.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged

Fixup 08cfb31465.

It is valid to use `null` or `false` in react children,
but the previous typings did not handle that case.

`ReactChild` is a subset of `ReactNode`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/98f1d85679fc581283cf6cc4d48ce8ae8b18e4a5/types/react/index.d.ts#L170

The commit is without tests for SVG, same as the commit being fixed.

### DIFF
--- a/test/should-fail.test.tsx
+++ b/test/should-fail.test.tsx
@@ -218,14 +218,6 @@ glamorous('div', {
   withProps: ''
 })
 
-glamorous('div', {
-  withProps: { visible: false }
-})(
-  (props) => ({
-    primaryColor: props.visible
-  })
-)
-
 glamorous('div')().withProps()
 glamorous('div')().withProps('')
 glamorous('div')().withProps(false)
--- a/typings/css-properties.d.ts
+++ b/typings/css-properties.d.ts
@@ -1926,7 +1926,7 @@ export interface CSSPropertiesLossy {
     | undefined
     | Array<CSSPropertiesComplete[keyof CSSPropertiesComplete]>
     | CSSPropertiesLossy
-    | React.ReactChild
+    | React.ReactNode
 }
 
 export interface CSSProperties
--- a/src/__tests__/__snapshots__/typescript.js.snap
+++ b/src/__tests__/__snapshots__/typescript.js.snap
@@ -75,44 +75,41 @@ test/should-fail.test.tsx(184,15): error TS2551: Property 'colors' does not exis
 test/should-fail.test.tsx(191,35): error TS2345: Argument of type 'StatelessComponent<ShouldClassNameUpdateProps>' is not assignable to parameter of type '\\"tspan\\"'.
 test/should-fail.test.tsx(207,17): error TS2551: Property 'colors' does not exist on type 'ShouldClassNameUpdateContext'. Did you mean 'color'?
 test/should-fail.test.tsx(217,11): error TS2345: Argument of type '\\"div\\"' is not assignable to parameter of type '\\"tspan\\"'.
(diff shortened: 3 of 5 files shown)


test/should-fail.test.tsx
```diff
@@ -218,14 +218,6 @@ glamorous('div', {
   withProps: ''
 })
```
luke-john: I'm assuming this was removed by accident as it looks like it should fail and appears unrelated to the changes. Would you mind re adding if that's the case thanks.

typings/css-properties.d.ts
```diff
@@ -1926,7 +1926,7 @@ export interface CSSPropertiesLossy {
     | undefined
     | Array<CSSPropertiesComplete[keyof CSSPropertiesComplete]>
     | CSSPropertiesLossy
-    | React.ReactChild
+    | React.ReactNode
```
luke-john: Thanks for the update, I'd love to remove this entirely from this file as a ReactNode/ReactChild is not a css property.

But keeping things working in the meantime sounds great 💯.

test/should-fail.test.tsx
denis-sokolov: This is removed on purpose, as this relaxation is the side-effect of the proposed change.

The new indexed type allows the value to be boolean, which makes the `primaryColor: props.visible` be valid.

Of course, it’s silly that we have this effect, but as noted in the other comment, it’d be awesome to not even have `children` in the `CSSProperties`. Until then this is a workaround.

The PR is effectively relaxing the type-checking for this instance, meaning fewer errors will be caught by TypeScript compiler. However, requirements that are too lax are much better than requirements that are too strict. In the latter case the developers fall into habit into silencing errors (with `as any`), defeating the entire purpose of typechecking in the first place.

typings/css-properties.d.ts
denis-sokolov: Me too, but small steps at a time. :-(

I have tried investigating the root cause, but I can not understand why does `children` match [the index signature on `CSSProperties`](https://github.com/denis-sokolov/glamorous/blob/ee948b134851ff9375a18a866fccc3c81d968488/typings/css-properties.d.ts#L1922) rather than the existing `children` on React components. I tried overriding `children` even directly in `CSSPropertiesLossy`, but the index still takes priority. By the way, if I remove `CSSPropertiesLossy` entirely, my code typechecks. But it, of course, might not be the solution.

luke-john:
Hey @denis-sokolov, sorry for the delay in looking at this. Really appreciate your work on this pr and following up on the issue.

I pulled this down for a play, and ended up creating a pr which solves the issue in a different way (removing the need for `React.ReactNode` in `CSSPropertiesLossy`).

https://github.com/paypal/glamorous/pull/372/files#diff-b6bac500e3d114b9016569ad1059fdd7
https://github.com/paypal/glamorous/pull/372

denis-sokolov:
If #372 works, clearly it is better than this hack. Let’s close this then, unless #372 does not work out, and then we’ll reopen this.

denis-sokolov:
Seeing how #372 is delayed, perhaps we could merge this in the meantime? (The build error is trivially fixable)

Ailrun:
@denis-sokolov I will merge #372 as soon as @luke-john updates snapshot. Sorry for delaying.

Ailrun:
#372 is merged. IMO, we could close this.

PR #370, opened 2018-01-04
fix(Contributing): remove outdated hook opt-in

denis-sokolov:
**What**: fixed documentation.

**Why**: because it was misleading.

**How**: by editing the text.

**Checklist**:

* [x] Documentation
* [ ] Tests N/A
* [x] Ready to be merged

  On a fresh clone and `npm install`, the hook was enabled already by `husky`, which [does not have opt-in](https://github.com/typicode/husky/issues/59).

Fixup 2a57031bcf.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,18 +49,6 @@ Please make sure to run the tests before you commit your changes. You can run
 `npm run test:update` which will update any snapshots that need updating.
 Make sure to include those changes (if they exist) in your commit.
 
-### opt into git hooks
-
-There are git hooks set up with this project that are automatically installed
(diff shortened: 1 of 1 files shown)


denis-sokolov:
Oooh, I see. I had the hook run and it was an annoyance for me, but the opting-in is not into the hook, but into one of the steps. Fair enough, sorry for the misunderstanding.

kentcdodds:
Oh, I see what you mean. Yeah, the hook will always run to do the formatting because that shouldn't affect most people as it rarely fails and ensures the PR has proper formatting. But I make the build/test/lint stuff opt-in only because I don't want to prevent people from submitting PRs if they're having trouble making the tests pass.

Issue #353, opened 2017-11-02
Typescript fails on Div and Button elements, if you give the onClick attribute a function as argument


Vanluren:
- `glamorous` version: 4.5.0
- `glamor` version: 2.20.40
- `react` version: 16.0.0
- `typescript` version: 2.6.1, 2.53

## Relevant code.

```javascript

<Button css={ buttonStyle } onClick={() => Raven.lastEventId() && Raven.showReportDialog()}>
File Report
</Button>

```

## What you did:

I tried to give a glamorous Div and a Button Element an onClick attribute with a Function as argument 

## What happened:

The following error is displayed:

```
Error:(30, 12) TS2322: Type '{ css: CSSProperties; onClick: () => void | ""; children: string; }' is not assignable to type 'IntrinsicAttributes & CSSProperties & ExtraGlamorousProps & HTMLProps<HTMLButtonElement> & { chil...'.
  Type '{ css: CSSProperties; onClick: () => void | ""; children: string; }' is not assignable to type 'CSSProperties'.
    Property 'onClick' is incompatible with index signature.
      Type '() => void | ""' is not assignable to type 'string | number | SingleOrArray<CSSPropertiesCompleteSingle, "alignContent" | "alignItems" | "ali...'.
        Type '() => void | ""' is not assignable to type 'ReactElement<any>'.
```

## Reproduction:

 I was not able to reproduce the problem using Stackblitz, but if you clone [this repo](https://github.com/Vanluren/latex-citation-generator)

The issue is present in the `src/index.tsx` file

## Problem description:

When the glamorous 'Button' element is given an the onClick attribute, typescript fails

## Suggested solution:

 If you change the CSSPropertiesLossy [glamorous/typings/css-properties.d.ts(l.1921)] interface to allow functions as a type, the error is not displayed, like so:

```javascript

export interface CSSPropertiesLossy {
  [propertyName: string]:
    | string
    | number
    | CSSPropertiesComplete
    | undefined
    | Array<CSSPropertiesComplete[keyof CSSPropertiesComplete]>
    | CSSPropertiesLossy
    | React.ReactChild
      | Function
}

```

Ailrun:
Hm... that's weird. I will investigate this.

Vanluren:
Any updtates on this?

Ailrun:
Actually,
``` typescript
const x: : CSSProperties | ExtraGlamorousProps | React.HTMLProps<HTMLButtonElement> {
    css: buttonStyle,
[...]

PR #372, opened 2018-01-08
fix(css-properties): remove incorrect React.ReactChild from Lossy Union

luke-john:
**What**:

Remove React.ReactChild from css properties Lossy Unions

This was originally added to fix a problem where built in glamorous components had been typed to accept the css properties union which included the lossy version.

By updating that component to instead use the css CompleteSingle interface the React.ReactChild is able to be removed.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged


denis-sokolov:
This clearly looks like a better solution. However, with TypeScript 2.6.1 your PR fails to typecheck my codebase.

```ts
<Div height={18}>
  <MyComponent />
</Div>
```

```
error TS2322: Type '{ height: 18; children: Element; }' is not assignable to type 'IntrinsicAttributes & CSSProperties & ExtraGlamorousProps & HTMLProps<HTMLDivElement> & { childre...'.
  Type '{ height: 18; children: Element; }' is not assignable to type 'CSSProperties'.
    Property 'children' is incompatible with index signature.
      Type 'Element' is not assignable to type 'string | number | SingleOrArray<CSSPropertiesCompleteSingle, "alignContent" | "alignItems" | "ali...'.
        Type 'Element' is not assignable to type '(string | number | {} | ("inherit" | "initial" | "unset" | "flex-start" | "flex-end" | "center" |...'.
```

I don’t understand why, since the test cases clearly include this. :-(

luke-john:
🤔 Are you able to share a version of `MyComponent` that is failing? I'm unable reproduce the issue locally (using 2.6.1 and 2.6.2).

I've added additional cases to test component signatures being accepted as children to this pr. https://github.com/luke-john/glamorous/blob/4a1598ca1a97d72c02ab54ecaccea3a1b9d8133c/test/glamorous.test.tsx#L568-L585

luke-john:
actually are you sure you're using this version? I'm a little bit suspicious of that error containing `CSSProperties`.

> `error TS2322: Type '{ height: 18; children: Element; }' is not assignable to type 'IntrinsicAttributes & CSSProperties & ExtraGlamorousProps & HTMLProps<HTMLDivElement> & { childre...'.`

denis-sokolov:
Hm, strange. I will test more thoroughly in the next day or two.

Ailrun:
@luke-john, @denis-sokolov is right.
Since the
```ts
{
  [propertyName: string]: T
}
```
overrides everything including `children`, it should include type for `children` or `ref`, `onClick` ...

typings/css-properties.d.ts
```diff
@@ -1926,7 +1926,6 @@ export interface CSSPropertiesLossy {
     | undefined
     | Array<CSSPropertiesComplete[keyof CSSPropertiesComplete]>
     | CSSPropertiesLossy
-    | React.ReactChild
```
Ailrun: This will cause the same problem of https://github.com/paypal/glamorous/issues/353

Ailrun: This is minimal PoC I'm able to imagine.
https://goo.gl/kjc8XD

Ailrun: IMO, to remove this extra union, we need to solve #353 first.

luke-john:
Ahh right, cheers for that @Ailrun, I've recently left my last workspace so don't have any large codebases to test prs on at the moment.

Something is clearly going wrong here as the type for glamorous.Div is not matching the type for the named export Div.

screen shot 2018-01-09 at 11 10 23 am

screen shot 2018-01-09 at 11 09 55 am

luke-john:
🤦‍♂️  Ahh I forgot to read the important comment at the top of that file

```
// The file `./named-built-in-glamorous-components.d.ts` is based off this file
// and should get any updates this file does.
```

will update shortly. thanks

typings/built-in-glamorous-components.d.ts
```diff
@@ -15,426 +16,670 @@ import {CSSProperties} from './css-properties'
...
   >
   Video: React.StatelessComponent<
-    CSSProperties & ExtraGlamorousProps & React.HTMLProps<HTMLVideoElement>
+    CSSPropertiesCompleteSingle &
+      ExtraGlamorousProps &
+      React.HTMLProps<HTMLVideoElement>
   >
   Wbr: React.StatelessComponent<
-    CSSProperties & ExtraGlamorousProps & React.HTMLProps<HTMLElement>
+    CSSPropertiesCompleteSingle &
```
Ailrun: So being unable to use `':active'` or other pseudo properties works as intended?
I mean,

```ts
const props = {
  ':active': {
    color: 'purple'
  }
};

<glamorous.A {...props} />
```

Is invalid?

luke-john: hmm, I hadn't really though about spreading the css props on.

🤔 That is valid, but as I understood it the purpose of that api was to make it possible to simple add css properties like follows;

```ts
<glamorous.A color="purple" />
```

And for the pattern you're describing do something like follows
```
const cssProps = {
  ':active': {
    color: 'purple'
  }
};
<glamorous.A css={props} />
```

I'm leaning towards making the types simpler here to reflect the intended use. As otherwise it makes the types less useful elsewhere.

Really appreciate your feedback and taking the time to go over these prs and even skimming the code 💖.

luke-john: We could add `CSSPropertiesPseudo` to the Intersection which would allow the case you've raised to work. But it would still fall down on things like

const props = {
  '> span': {
    color: 'purple'
  }
};

<glamorous.A {...props}>hello <span>world</span></glamorous.A>

Ailrun: Thank yo for your kind answer!
I think I fully get the point what you want to say, and I agree with it to some distance. I will approve PR.

luke-john: I've added test cases for the pseudo props as part of this pr as well, cheers.

https://github.com/luke-john/glamorous/blob/a821469f2cad3c8511bbdfc12d2c807d6cc4d08f/test/glamorous.test.tsx#L611-L621

Ailrun:
@luke-john Following is just a question. If you give me an answer, I would greatly appreciate it.

Is there any reason for dismissing the review?

denis-sokolov:
My speculation would be Luke wanted to express a need for you to review it afresh after the latest changes, @Ailrun.

I promised to test this again until today, @luke-john, but my schedule can’t allow it, apologies. I won’t be able to reproduce it until the weekend.

luke-john:
@Ailrun I pushed the `CSSPropertiesPseudo` and test case additions for  https://github.com/paypal/glamorous/pull/372#discussion_r160320700 after your review.

And GitHub automatically dismisses reviews when changes are made to a pull request.

Ailrun:
@luke-john Thank you for explanation! I didn't know there's a case that github automatically dismisses a review.
Anyway, in my point of view, it looks OK.

Ailrun:
And also thank you for @denis-sokolov to try to explain what I ask! I really appreciate you all!

denis-sokolov:
GitHub does not always automatically dismiss a review, it’s a per-repository setting. Some people want to make sure that the reviewers have seen the absolutely latest version of the PR, and some are okay with relying on a more lax rule of common sense.

denis-sokolov:
I’m confused by my tests. Minimal case that triggers the error with this branch as the dependency is this:

```tsx
<Div>
  <Div />
</Div>
```

Obviously it is supposed to be covered in tests, but I was pretty thorough with the tests. Building `glamorous` at 8a0216f makes my `tsc` succeed, then I switch glamorous to 7912617, and `tsc` starts loudly failing.

This happens with tsc 2.6.1 and 2.6.2, confirmed the flow with both.

Here is the error:

```
index.tsx(90,7): error TS2322: Type '{ children: Element; }' is not assignable to type 'IntrinsicAttributes & CSSProperties & ExtraGlamorousProps & HTMLProps<HTMLDivElement> & { childre...'.
  Type '{ children: Element; }' is not assignable to type 'CSSProperties'.
    Property 'children' is incompatible with index signature.
      Type 'Element' is not assignable to type 'string | number | SingleOrArray<CSSPropertiesCompleteSingle, "alignContent" | "alignItems" | "ali...'.
        Type 'Element' is not assignable to type '(string | number | {} | ("inherit" | "initial" | "unset" | "flex-start" | "flex-end" | "center" |...'.```

TeffenEllis:
@Ailrun,

This looks awesome! Is there an ETA to release?

Ailrun:
@TeffenEllis Sorry for late response. @kentcdodds could you answer to @TeffenEllis ?

kentcdodds:
The releases for this project are automatic. So it should be released as soon as this is merged.

Ailrun:
@kentcdodds Oh, I mean... when will this merged?

kentcdodds:
I keep out of typescript related stuff unless specifically asked. Do you all my help here?

Feel free to merge typescript pull requests yourself. Just make sure to follow the commit message convention when doing squash and merge so the right version is released.

Ailrun:
@kentcdodds I got 
image this message so I thought I can't merge this.
Am I wrong? Can I merge this?

kentcdodds:
It's because the PR was not updated with `master`, so I just clicked the "Update branch" button and once the status checks are done you should be able to merge it.

Just make sure to follow [the commit message conventions](https://github.com/paypal/glamorous/blob/master/other/MAINTAINING.md) when you squash and merge, otherwise a new version will not be released.

Cheers! And thanks!

Ailrun:
@luke-john Please update your snapshots.

luke-john:
Have just updated the tests, I'm unsure about merging this though as we never worked out why this was causing additional failures https://github.com/paypal/glamorous/pull/372#issuecomment-359251320.  I've tried to reproduce without luck, however have added a [test case](https://github.com/paypal/glamorous/pull/372/commits/8738c766f1efd1948499540b5ddac43ee857d4e5) for this which does not appear to fail.

Also sorry for the huge delay here, I've just relocated to London from Australia and my life has been a bit hectic lately.

luke-john:
thanks for following this up @Ailrun

Ailrun:
@luke-john it looks like @denis-sokolov miss something (or we miss some updates) since error message includes `CssProperties` which we just removed in this PR. I will also retry to reproduce, and if it's failed, IMO, we should merge this first and make an issue with `needs investigation` label.

Ailrun:
@kentcdodds I still cannot merge this.
image
There're no buttons I can click :(.

Issue #376, opened 2018-01-18
Components returned from `withComponent` use the wrong rootEl


tikotzky:
* `glamorous` version: 4.11.3
* `glamor` version: 2.20.40
* `react` version: 16.2.0

Relevant code.

```javascript
const A = glamorous.div({}).withComponent('a');

function Link() {
	<A href="https://www.google.com">href is missing 🙁</A>
}
```

What you did:
Tried to render an `a` tag with an `href` attribute

What happened:
It rendered an `a` tag without the `href` attribute

Reproduction:
https://codesandbox.io/s/vvzq1kwlw0

Problem description:
This broke in 4.11.3
Here is a codesandbox with the same code working on 4.11.2
https://codesandbox.io/s/3xr58oxqv6

Suggested solution:
Dont know the internals well enough 😄

kentcdodds:
Hi @tikotzky!

Thanks for the issue. It probably broke when we merged and released #374 (cc @exah).

@tikotzky, do you think you could write a unit test to reproduce this issue? Probably in this file: https://github.com/paypal/glamorous/blob/master/src/__tests__/create-glamorous.with-component.js

If you feel so inclined to provide the fix as well that would be welcome :)

Thanks!

tikotzky:
I already have a unit test which reproduces the issue :)
I'll open a PR in a few minutes.

I'll try to fix but its my first time peering under the hood at glamorous, so might take a few...

kentcdodds:
Hopefully looking at #374 will give you some clues. Thanks!

tikotzky:
I opened a PR with a failing test here https://github.com/paypal/glamorous/pull/377

After a little more testing it seems like the test passes on 4.11.2, and starts failing after
https://github.com/paypal/glamorous/commit/d9560fa6ac9afcdd214ce211f8533280beb7062c.

Judging from that it seems like a dep broke it.
I'll dig a little further to see if I can figure it out.

kentcdodds:
Thanks!

tikotzky:
The test seemed to be breaking due to a change in enzyme...
I fixed the test and have a fix...
Gonna update the PR soon.

tikotzky:
Pr updated.

Thanks for the awesome lib 😄 👍

kentcdodds:
Fixed!

PR #377, opened 2018-01-18
Reset roolEl when calling withComponent

tikotzky:
**What**:
A fix and failing test for https://github.com/paypal/glamorous/issues/376

**Why**:
Added as per https://github.com/paypal/glamorous/issues/376#issuecomment-358795380

**How**:
When calling `withComponent` the `rootEl` is now set to the component being passed in.
You can still manually set the `rootEl` by passing in `options` in the second param.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged  


src/create-glamorous.js
```diff
@@ -114,6 +114,7 @@ function createGlamorous(splitProps) {
           {
             ...componentProperties,
             comp: newComp,
+            rootEl: newComp,
```
kentcdodds: This should probably be `rootEl: getRootEl(newComp)` just in case someone passes `withComponent(SomeGlamorousComponent)` :+1:

tikotzky: done

tikotzky:
@kentcdodds one other thing...
[this test](https://github.com/paypal/glamorous/blob/master/src/__tests__/index.with-props.js#L121-L127) currently is not testing anything since `.find('div')` is not finding the div.
This means that this test will always pass even if the props are getting set.

I can fix it by removing `.find('div')`, but the better fix would be to switch to a snapshot test as well, this way you would not have to worry about it wrongly being undefined in the future.
Should I switch my test and the above test to use snapshots?

src/create-glamorous.js
marzelin: hmm... `getRootEl` might do more harm than help there. Could you try if this example works for your code? (bigger font size should be applied) https://codesandbox.io/s/pkr1rn24vx

kentcdodds: Hmm... I'm not sure I understand. I think your codesandbox simply reveals the bug that would be fixed by these changes.

kentcdodds:
Thanks for catching that. Let's go ahead and merge this and you can fix that test in another PR if you like. I like to try to avoid snapshots if I can, preferring a more explicit assertion. So if you're able to come up with something more explicit it would be appreciated. Thanks!

kentcdodds:
Thanks so much for your help @tikotzky! I've added you as a collaborator on the project. Please make sure that you review the `other/MAINTAINING.md` and `CONTRIBUTING.md` files (specifically the bit about the commit messages and the git hooks) and familiarize yourself with the code of conduct (we're using [the contributor covenant](http://contributor-covenant.org/)). You might also want to [watch the repo](https://help.github.com/articles/watching-repositories/) to be notified when someone files an issue/PR. Please continue to make PRs as you feel the need (you can make your branches directly on the repo rather than your fork if you want). Thanks! And welcome to the team :)

src/create-glamorous.js
marzelin: @kentcdodds I didn't have time to explain this properly yesterday, so here it goes:

Normally, when you pass a glamorous component to a `glamourous()`, it 'merges' all the glamorous props so that there is only one instance of a glamorous component. `withComponent` works differently - when you pass a glamorous component to it there will be two instances - one glamorous instance rendering the other.

`rootEl` purpose is to decide what props should be filtered out (props that are invalid for a given `rootEl` won't be passed to the rendered component). When you assign `rootEl` to a wrapped non-glamorous component (mosty html element) all custom props will be filtered out.

This causes problems when this inner glamorous component (the one given to `withComponent`) relies on custom props for dynamic styles (it won't be able to access the custom props).

In summary: when passing glamorous component to `withComponent`, the component that will be directly rendered is the glamorous component itself so it should be assigned to `rootEl`.

tldr;
this works in 4.11.2 https://codesandbox.io/s/pkr1rn24vx
but doesn't in the latest (4.11.4)  https://codesandbox.io/s/n0yrxyrnl

kentcdodds: I think I understand. Could you perhaps open a pull request which adds a test for your scenario and see if we can satisfy both cases? If not then we can have a discussion about which to favor...

PR #378, opened 2018-01-22
Fix multiple issues with jest example

penx:
**What**: Fixed multiple issues with the jest example

**Why**: Because the following instructions for running the Jest example no longer work:

```
git clone https://github.com/paypal/glamorous.git
cd glamorous/examples/with-jest
npm install
npm run test
```

Fix several issues occur during `npm test`:

1.

>     react-dom is an implicit dependency in order to support react@0.13-14. Please add the appropriate version to your devDependencies. See https://github.com/airbnb/enzyme#installation

So added react-dom as a devDependency

2.

> react-dom@15.5+ and react-test-renderer are implicit dependencies when using react@15.5+ with enzyme. Please add the appropriate version to your devDependencies. See https://github.com/airbnb/enzyme#installation

So added react-test-renderer as a devDependency and updated versions of React to 15.5+

3.

>     SyntaxError: Unexpected token import

So added .babelrc with es2015 preset

4.

> Unexpected token (8:4)
>          6 | test('enzyme', () => {
>          7 |   const ui = (
>       >  8 |     <Wrapper>
>            |     ^

So added react to babel presets

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged 


kentcdodds:
Thanks @penx!

While we're updating things, could you just update everything to the latest? (so rather than react@15, let's use react@16). Also, instead of `babel-preset-es2015` (deprecated), let's use `babel-preset-env`. Thanks!

penx:
done other than updating to React@16, as this causes `npm test` to throw an error:

>     TypeError: Cannot read property 'string' of undefined
> 
>       at Function.glamorousComponentFactory [as a] (node_modules/glamorous/dist/glamorous.cjs.js:206:28)
>           at Array.reduce (<anonymous>)
>       at Object.<anonymous> (index.js:1:1065)
>       at Object.<anonymous> (index.test.js:4:14)
>           at <anonymous>
>       at process._tickCallback (internal/process/next_tick.js:188:7)

kentcdodds:
I'll go ahead and merge this. If you'd like to dig in deeper to figure out what that problem is that'd be great, but either way is fine :)

Thanks so much!

PR #388, opened 2018-02-21
tech(node): add node as a dep.

### LINKED ISSUE #352: `npm run validate` fails on current master
I just checked out the current master and followed the steps from the [CONTRIBUTING.md](https://github.com/paypal/glamorous/blob/master/CONTRIBUTING.md).

Calling `npm run validate` lead to [`src/__tests__/theme-provider.js`](https://github.com/paypal/glamorous/blob/0cc10069dc95c7f530435715d4c4ef4135c989ae/src/__tests__/theme-provider.js) failing:

screen shot 2017-10-26 at 14 02 19

I guess it's due to some dependency not following semver too strictly and now it's representing Components differently.

But then again I'm not too sure.

Given the big amount of dependencies, I'm surprised there's no yarn or npm5 lockfile for glamorous.
As far as I understand, lockfiles within `/node_modules/` [are ignored](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) (at least with yarn). So having them won't harm users of glamorous, but it'll benefit contributors.

kentcdodds:
**What**: Adds node as a dependency. This is a brilliant idea from the folks at npm. It allows us to have any version of node and as long as the folks using the project only use npm scripts to run our stuff, the installed version of node will be the one used!

**Why**: Closes #352

**How**: add node to the dev dependencies and update contributing docs to be a little simpler.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged 
* [ ] Added myself to contributors table N/A


CONTRIBUTING.md
```diff
@@ -8,9 +8,8 @@ Thanks for being willing to contribute!
 ## Project setup
 
 1. Fork and clone the repo
-2. `$ npm install` to install dependencies
-3. `$ npm run validate` to validate you've got it working
-4. Create a branch for your PR
+2. `npm run setup --silent` to setup the project (it installs deps and runs the validate script)
```
luke-john: I don't think the `--silent` is needed here, that's already been added to the validate command under the setup script.

package.json
```diff
@@ -56,6 +57,7 @@
     "glamor": "^2.20.37",
     "jest-glamor-react": "^3.2.2",
     "kcd-scripts": "^0.31.1",
+    "node": "^8.9.4",
```
luke-john: Great idea adding this, hopefully it should make contributing a bit easier and cut down on issues around node versions 🙂.

CONTRIBUTING.md
kentcdodds: Nope, still need it to prevent npm from logging useless stuff when running the command I'm afraid.

ajwhite:
That's really cool

> This is a brilliant idea from the folks at npm.

Is there a source you could share? Tried to find more information about this, but it's a tricky search term to research 😛

kentcdodds:
This is as official as I could get: https://twitter.com/maybekatz/status/958157474397171712

It's not an npm thing actually. It's just a normal npm package! Pretty interesting exercise to go through the code actually if you're curious. Start here: https://unpkg.com/node/

ajwhite:
Forgot to follow up and mention the README of https://www.npmjs.com/package/node pretty much sums it up!

PR #397, opened 2018-03-16
Failing test and proposed fix for issue #396

### LINKED ISSUE #396: Bug: unit test throws "rule cache miss occurred" when nesting components
* `glamorous` version: 4.12.1
* `glamor` version: 2.20.40
* `react` version: 16.2.0

# Relevant code:

## Component
`Example.js`
```javascript
import glamorous from "glamorous";
import React from "react";

var InnerOne = glamorous.div({
  color: '#F00'
});

const InnerTwo = props => <InnerOne {...props} />;

const Example = glamorous(InnerTwo)(({ color }) => ({
  color: color ? color : undefined
}));

export default Example;
```

## Test
`Example.test.js`
```javascript
import React from "react";
import ReactDOM from "react-dom";
import Example from "./Example";

describe("Component - Example", () => {
  it("renders without crashing", () => {
    const div = document.createElement("div");
    ReactDOM.render(<Example type="text" />, div);
  });
});
```

# What you did:

1. Compose a component from another Glamorous component that has a JSX wrapper.
2. Try to unit test this component

# What happened:

The component will render fine in storybook etc. but when we run the test we get an error:

```
    [glamor] an unexpected rule cache miss occurred. This is probably a sign of multiple glamor instances in your app. See https://github.com/threepointone/glamor/issues/79
```

We've also tried with another test framework and get the same error.

# Reproduction:

Shown in PR with breaking unit test - https://github.com/paypal/glamorous/pull/397

penx:
Fixes #396

### DIFF
--- a/src/__tests__/__snapshots__/index.js.snap
+++ b/src/__tests__/__snapshots__/index.js.snap
@@ -75,6 +75,27 @@ exports[`can create custom built-in glamorous components 1`] = `
 />
 `;
 
+exports[`can render nested components 1`] = `
+.glamor-0,
+[data-glamor-0] {
+  color: #F00;
+}
+
+<glamorous(unknown)>
+  <Component
+    className=""
+  >
+    <glamorous(div)
+      className=""
+    >
+      <div
+        className="glamor-0"
+      />
+    </glamorous(div)>
+  </Component>
+</glamorous(unknown)>
+`;
+
 exports[`can use pre-built glamorous components with css attributes 1`] = `
 .glamor-0,
 [data-glamor-0] {
@@ -134,7 +155,7 @@ exports[`forwards props when the GlamorousComponent.rootEl is known 1`] = `
 
 exports[`merges composed component forwardProps 1`] = `
 <div
-  class="css-nil"
+  class=""
 />
 `;
 
@@ -294,7 +315,7 @@ exports[`the css prop accepts "GlamorousStyles": css prop with an object 1`] = `
 
 exports[`will forward \`color\` to an svg 1`] = `
 <svg
-  class="css-nil"
+  class=""
   color="red"
 />
 `;
--- a/src/__tests__/index.js
+++ b/src/__tests__/index.js
@@ -404,3 +404,17 @@ test('should accept user defined contextTypes', () => {
   const props = {theme}
   expect(dynamicStyles).toHaveBeenCalledWith(props, context)
 })
+
+test('can render nested components', () => {
+  const InnerOne = glamorous.div({
+    color: '#F00',
+  })
+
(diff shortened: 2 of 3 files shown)


kentcdodds:
Interesting! Thank you for this! Do you think you could dig in a little bit and figure out what is up?

penx:
I intend to when I have some time 🙂 hopefully next week. Have you seen similar issues previously? Any pointers to bits of code/files in Glamorous that could be relevant and would be worth reading or stepping through with a debugger attached?

kentcdodds:
Awesome :+1:

No, I've never seen this issue before. It's surprising to me that we don't already incidentally have a test for this. As for a good place to look, I honestly have no idea. This feels like a integration with glamor issue, you'll find our use of the glamor API here:

https://github.com/paypal/glamorous/blob/29c5afde78380eff6a0002d2f2e8dd8c8c5c6f50/src/get-glamor-classname.js#L65

Thanks! And good luck!

penx:
I've had a chance to run through this with a debugger but not figured out everything that's happening yet. Progress so far:

1.

https://github.com/paypal/glamorous/blob/29c5afde78380eff6a0002d2f2e8dd8c8c5c6f50/src/get-glamor-classname.js#L65

In the failing unit test, this is called with `devRules=null` and

```js
mappedArgs=[{
  color: '#F00'
}, {
  'data-css-nil': ''
}, {
}, undefined
]
```

2. 

https://github.com/threepointone/glamor/blob/db533d7c5fe249e443756367161d0830086b0624/src/index.js#L360-L365

`isLikeRule({data-css-nil: ""})` returns true and `_getRegistered(_src)` is called

3.

https://github.com/threepointone/glamor/blob/db533d7c5fe249e443756367161d0830086b0624/src/index.js#L265-L274

`idFor({data-css-nil: ""})` returns "nil"

`var ret = registered[idFor(rule)]` returns null and the error is thrown

penx:
@kentcdodds I've created a failing upstream test on Glamor, perhaps you an idea on what's going on here before I continue to debug?

https://github.com/threepointone/glamor/pull/373

penx:
Further to the above:

https://github.com/paypal/glamorous/blob/29c5afde78380eff6a0002d2f2e8dd8c8c5c6f50/src/get-glamor-classname.js#L65

This is initially called with:

```
devRules=null
mappedArgs=[{},null,{},null]
```

Which returns `css-nil`.

This value is later used as the CSS class name of the element when this function is called a second time.

penx:
I've added a potential fix to this PR that ignores the css class returned from glamor if it's 'css-nil'.

kentcdodds:
This project is now in an unmaintained status. Please see the README for more information.

PR #409, opened 2018-04-09
fix(innerRef): allow ref as object

viktorivarsson:
React 16.3 comes with a createRef API where you can add ref as an object

**What**:
Allow innerRef to be passed as an object.

**Why**:
Because React 16.3 comes with a createRef API, the ref to elements is not only functions anymore, it may also be objects.

**How**:
I have updated prop type for the innerRef prop as well as the TypeScript interface.

**Checklist**:

* [ ] Documentation N/A
* [ ] Tests N/A
* [x] Ready to be merged

I have updated where I can see that we set types for innerRef, but it would be great if someone took a quick look as well.

I'm happy to update if there are adjustments needed.


kentcdodds:
Ah, looks like there's an issue with the snapshots. Could you run `npm run test:update` and commit and push the changes? Thanks!

viktorivarsson:
Just glad I was able to help! Snapshots are now updated.

PR #411, opened 2018-04-13
Use csstype for standardized CSS typings

### LINKED ISSUE #408: Use csstype
[`csstype`](https://github.com/frenic/csstype) is a set of standard, high-quality CSS typings generated from MDN data. It is now being used by the React and JSS typings. It would be great to have the entire CSS-in-JS ecosystem standardize on these typings rather than each project maintaining its own pet CSS typings, each with varying degrees of completeness and usually inconsistent with one another!

pelotom:
Resolves #408. See that issue for details and rationale.


typings/svg-properties.d.ts
```diff
@@ -1,308 +1,8 @@
...
-  xmlSpace?: string
-  y1?: number | string
-  y2?: number | string
-  y?: number | string
-  yChannelSelector?: string
-  z?: number | string
-  zoomAndPan?: string
-}
+export interface SVGPropertiesCompleteSingle
+  extends CSS.SvgProperties<number | string> {}
```
tikotzky: should this be `CSS.SvgPropertiesFallback`?
It seems like it would need to be in order to allow using arrays to set fallback values.

tikotzky: It does look like it was not supporting array until now, though i'm pretty sure that glamor supports it...

pelotom: I would have used `CSS.SvgPropertiesFallback` if I were writing these typings from scratch, but to minimize breakage I was trying to keep as many of the existing exports intact as possible. So this is accomplishing what `CSS.SvgPropertiesFallback` would:

```ts
type SVGPropertiesComplete = SingleOrArray<
  SVGPropertiesCompleteSingle,
  keyof SVGPropertiesCompleteSingle
>
```

pelotom:
I'm not sure what all the test failures are about, but these:

```
[test:cover]     - test/should-fail.test.tsx(289,35): error TS2322: Type '{ display: "blocks"; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<object & CSSProperties & ExtraGlamorousP...'.
[test:cover]     -   Type '{ display: "blocks"; }' is not assignable to type 'Readonly<object & CSSProperties & ExtraGlamorousProps>'.
[test:cover]     -     Types of property 'display' are incompatible.
[test:cover]     -       Type '"blocks"' is not assignable to type '"none" | "ruby" | "table" | "inherit" | "initial" | "unset" | "contents" | "block" | "inline" | "...'.
```

```
[test:cover]     - test/should-fail.test.tsx(293,29): error TS2322: Type '{ display: "blocks"; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<HTMLProps<HTMLDivElement> & object & CSS...'.
[test:cover]     -   Type '{ display: "blocks"; }' is not assignable to type 'Readonly<HTMLProps<HTMLDivElement> & object & CSSProperties & ExtraGlamorousProps>'.
[test:cover]     -     Types of property 'display' are incompatible.
[test:cover]     -       Type '"blocks"' is not assignable to type '"none" | "ruby" | "table" | "inherit" | "initial" | "unset" | "contents" | "block" | "inline" | "...'.
```

should be resolved by https://github.com/frenic/csstype/issues/8. In the meantime, it allows a `string` because otherwise you forbid valid multi-part values.

Ailrun:
I think you should update negative tests and  snapshots.

src/__tests__/__snapshots__/typescript.js.snap
```diff
@@ -38,13 +38,7 @@ test/should-fail.test.tsx(100,24): error TS2551: Property 'colors' does not exis
 test/should-fail.test.tsx(111,3): error TS2344: Type 'PropsWithoutTheme' does not satisfy the constraint '{ theme: any; }'.
   Property 'theme' is missing in type 'PropsWithoutTheme'.
 test/should-fail.test.tsx(119,3): error TS2345: Argument of type 'StatelessComponent<object>' is not assignable to parameter of type '\\"tspan\\"'.
-test/should-fail.test.tsx(134,3): error TS2345: Argument of type '(props: { theme: any; } & ExampleComponentProps & object) => { display: \\"none\\" | \\"hidden\\"; }' is not assignable to parameter of type 'StyleArgument<CSSProperties, { theme: any; } & ExampleComponentProps & object>'.
```
Ailrun: Those are in should-fail tests, so I think you should remove relevant tests too. I mean, line 134 in test/should-fail.test.tsx.

pelotom: These tests will be valid again once https://github.com/frenic/csstype/issues/8 is merged. Would it make sense to just add comments to the effect that the lines are only temporarily not being flagged with errors, pending that PR?

Ailrun: @pelotom Oh, sorry. I though https://github.com/frenic/csstype/issues/8#issuecomment-364845647 says that '`string` is valid', but it does not... Sorry for giving confusion.

Ailrun:
Looks OK for me.

kentcdodds:
Thanks folks! This is great to remove so much code!

PR #416, opened 2018-05-01
Add typescript typings for preact

aaronjensen:
**What**:

These are working typescript types for Preact. What doesn't work is the build.
To make it work, the `preact/package.json` would need to have an entry for
typings and the typings would need to be copied to the `preact` folder. Since
all of this seems it would need to happen in `kcd-scripts` maybe @kentcdodds has
some time to help with that?

``` javascript
{
  "main": "dist/glamorous.cjs.js",
  "jsnext:main": "dist/glamorous.esm.js",
  "module": "dist/glamorous.esm.js",
  "typings": "typings/glamorous.d.ts"
}
```

**Why**:

There are currently no typescript types for preact

**How**:

I upgraded a medium sized preact-glamorous project.

**Checklist**:

* [ ] Documentation N/A
* [x] Tests
* [x] Ready to be merged 


Ailrun:
Thank you for your pull request!
It looks like what you did is
1. Copy-and-paste original types
2. Change `react` to `preact`
3. Change `react` only types(`React.Component`...) to `preact` types(`preact.Component`...)
4. Add `IntrinsicAttributes` and `IntrinsicClassAttributes` in the place of `JSX.{something}`.
Am I right? Are there anything important I missed?
If this is the case, it looks like the only thing this PR requires is adding some tests... of course, after fixing typing-loading issues.

kentcdodds:
> What doesn't work is the build.

I don't think I want to bake that into kcd-scripts just yet. So let's add a `postbuild` step that calls into a node script to modify the generated `preact/package.json`. Should be just three or four lines.

PR #425, opened 2018-05-31
Fix typings for Typescript 2.9

aaronjensen:


typings/preact/helpers.d.ts
```diff
@@ -1,5 +1,7 @@
-type Diff<T extends string, U extends string> = ({[P in T]: P} &
-  {[P in U]: never} & {[x: string]: never})[T]
+type Diff<
+  T extends string | number | symbol,
+  U extends string | number | symbol
+> = ({[P in T]: P} & {[P in U]: never} & {[x: string]: never})[T]
```
Ailrun: However, after fix this, Glamorous typing will not work with TS <= 2.8

aaronjensen: Ah, have you tested that? I'm guessing the `in T` and `in U` don't work? this is why flow's approach is better (versioning to lib and flow instead of just lib). Any way, it seems like we should prefer supporting the most recent version of TS, no?

Ailrun: @aaronjensen As you said, supporting recent TS version is better. However, if there's a way to support both old and new versions, that's the best, isn't it?
I mean, `keyof any` will return `string | number | symbol` for TS@2.9, and `string` for TS@<=2.8, so how about that?

Ailrun: And if we want to support TS@2.9 only, we have `Extract` from TS@2.8, so we don't need these helpers.

aaronjensen:
@Ailrun I used `Exclude` since that's 2.8+

Ailrun:
@aaronjensen So for you `T extends keyof any` looks weird? That's why you remove support for TS@>=2.6.2 < 2.8?

kentcdodds:
Can we get the build working before we merge this please? :)

aaronjensen:
We could yeah, but it has nothing to do w/ this PR, so should it be a separate one? This repo doesn't appear to use a package-lock or a yarn.lock so as new eslint rules get added failures crop up with 0 changes.

Current failures:

```
[lint] /home/travis/build/paypal/glamorous/src/theme-provider.js
[lint]   16:1  error  componentWillMount is deprecated since React 16.3.0, use UNSAFE_componentWillMount instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillmount                       react/no-deprecated
[lint]   16:1  error  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops  react/no-deprecated
[lint] 
[lint] /home/travis/build/paypal/glamorous/src/with-theme.js
[lint]   16:3  error  componentWillMount is deprecated since React 16.3.0, use UNSAFE_componentWillMount instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillmount                       react/no-deprecated
[lint]   16:3  error  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops  react/no-deprecated
```

aaronjensen:
I actually get even more lint failures locally 😖

aaronjensen:
Tests are passing now with the deprecated lints being ignored. Maybe a follow-up PR could deal w/ it properly (which would probably require bumping the minimum version of react required)

package.json
```diff
@@ -71,7 +71,7 @@
     "tslint": "^5.8.0",
     "tslint-microsoft-contrib": "^5.0.0",
     "tsutils": "^2.15.0",
-    "typescript": "^2.6.2"
+    "typescript": "^2.9.1"
```
Ailrun: Does not this work with TS@2.8?

aaronjensen: The snapshot tests won't. This is just a dev dependency.

Ailrun: Oh, snapshot... OK.

Issue #419, opened 2018-05-09
Deprecate glamorous 💄 in favor of emotion 👩‍🎤


kentcdodds:
For a long time now I've been considering deprecating glamorous (NOT glamorous-native) in favor of [emotion](https://emotion.sh/).

**Here's my reasoning**:

1. Emotion can do everything glamorous can do
2. Emotion can do more than glamorous can do
3. Emotion is faster than glamorous
4. Emotion is smaller than glamorous
5. It's good for the community to consolidate and rally around using and improving a single solution

I would not recommend people spend a bunch of time migrating to emotion. glamorous isn't bad, it's just not as good as emotion and I don't think it would be worth the effort to manually migrate. That said, I'm fairly confident that a codemod could be written to automatically migrate your glamorous code to emotion.

**What a deprecation would mean**:

- I would add a notice to the README and website indicating that this project is no longer maintained and encouraging people to use emotion instead.

**What a deprecation would NOT mean**:

- glamorous-native is still the best-in-class for styling native applications and is entirely separate from glamorous. It would NOT be deprecated. It is an entirely separate project with no shared dependencies anyway, so a deprecation of glamorous wouldn't impact glamorous-native directly.

**What I want from you**:

Please give this a :+1: if you agree or a :-1: if you disagree.

**If you disagree, please list reasons why and whether you would be willing to assume the maintainer role.**

Thanks friends!

kentcdodds:
Things are looking very positive for this. If someone would like to help a LOT of people here are a few things you could do:

1. Write a migration guide. It could be a simple markdown file [here](https://github.com/paypal/glamorous/tree/master/other).
2. Write a codemod. If you don't have any experience with ASTs then this might be a bit of a challenge for you. I have some resources for you though:

* [Writing custom Babel and ESLint plugins with ASTs](https://youtu.be/VBscbcm2Mok?list=PLV5CVI1eNcJgNqzNwcs4UKrlJdhfDjshf): A 53 minute talk by [@kentcdodds](https://twitter.com/kentcdodds)
* [babel-handbook](https://github.com/thejameskyle/babel-handbook): A guided handbook on how to use Babel and how to create plugins for Babel by [@thejameskyle](https://twitter.com/thejameskyle)
* [Code Transformation and Linting](https://kentcdodds.com/workshops/#code-transformation-and-linting): A workshop (recording available on Frontend Masters) with exercises of making custom Babel and ESLint plugins

[...]

PR #428, opened 2018-06-17
Add codemod that migrates glamorous to emotion

TejasQ:
In response to #419, I added a link to a codemod I wrote that helps migrate existing React/Preact codebases to emotion from glamorous.


TejasQ:
For sure!

